Issue #9, opened 2020-07-14
Design the structure of the Redux store


mscharfstein:
Design the structure of the Redux store:
- Includes the different slices of reducer logic/actions for the Redux state (in our case, probably TWO slices: one with the questions (populated on initial load), and one with all the answers)
- The slices have reducers, which can have different functions / actions that update the state (in our case, for the answers reducer, probably just an "answerSelected" action (maybe separate for boolean and numeric answers))
- Make sure to export the reducers
- Assumption: Components can look up individually the questions based on their IDs

Reducer for answers can look similar to:

```
const answersSlice = createSlice({
  name: 'answers',
  initialState,
  reducers: {
   answerSelected(state, action) {
      // define the structure of `action.payload`
      // should support both boolean & numeric answers (separate reducers, maybe?)
      // must also have an identifier to lookup and modify only the corresponding
      // question
    }
  }
})
```

Separate reducer slice for questions.

Notes:
- Consider first submitting a PR to create the redux state with an initial state before working on the reducer(s) - this will unblock Issue #10, Issue #12 
- After unblocking the other issues, implement the reducer(s)

kebrahim:
this should also include slices for the form, sections & services data from the API. see https://github.com/ny/maslow/blob/master/server/src/main/resources/application.yaml

PR #26, opened 2020-07-15
Add answers reducer with an answerUpdate action

mscharfstein:
Works on #9 

Creates an "answers" slice - includes an answers object within the state (the object will contain keys representing questions, and values representing answers), an action to update an answer, and a reducer to handle this action. The action will contain a payload with both the question key and the answer value. These fields will eventually be passed based on the information filled out in the form.

Note: I did NOT specify an initial state, though we had discussed doing so. The initial state should really be the empty object. The reducer is already set up so the state can be updated, and even if it is not, this should not be blocking because the app should be able to handle the case where the state is null.

The counters reducer will be removed in a following PR, as will the questions reducer be added. We don't have any accessibility checks set up yet but this is just code in the background so it shouldn't matter.


client/src/features/answers/answersSlice.js
```diff
@@ -0,0 +1,22 @@
+import { createSlice } from "@reduxjs/toolkit";
+
+export const answersSlice = createSlice({
+  name: "answers",
+  initialState: {
+    // will include answers for questions including:
+    // householdSize, adults65AndOlder, children13through17, children6through12,
```
mscharfstein, line 7: should these "keys" be included in the data we store? or should the FE just store them as constants and determine which one should be used upon a question's answer? i'd prefer the former unless there's some reason against it. @marcja since I know you had thoughts about this

client/src/app/store.js
```diff
@@ -1,8 +1,10 @@
 import { configureStore } from "@reduxjs/toolkit";
 import counterReducer from "../features/counter/counterSlice";
+import answersReducer from "../features/answers/answersSlice";
```
marcja, line 3: s/answers/responses/g?

client/src/features/answers/answersSlice.js
marcja: That is correct. See #28 for details.

client/src/features/answers/answersSlice.js
```diff
@@ -0,0 +1,22 @@
...
+    // children5orYounger, monthlyHouseholdIncome, employed, student, military, disabled
+    // but should start empty
+  },
+  reducers: {
+    answerUpdate: (state, action) => {
+      state[action.payload.question] = action.payload.answer;
```
marcja, line 13: s/question/questionKey/g?

egracer: +1 to prevent confusion esp. since `questions` will be a different slice

client/src/features/answers/answersSlice.test.js
```diff
@@ -0,0 +1,40 @@
+import "@testing-library/jest-dom/extend-expect";
+import answers, { answerUpdate } from "./answersSlice";
```
egracer, line 2: I assume that `answers` is Redux Toolkit magic, since this variable isn't exported in the `answersSlice` file?

client/src/features/answers/answersSlice.test.js
```diff
@@ -0,0 +1,40 @@
...
+
+test("initial state", () => {
+  expect(answers(undefined, {})).toEqual({});
+});
+
+test("answer update - integer", () => {
```
egracer, line 8: Called 'update' but performs an add. Is it worth explicitly testing both cases? Or is that just re-testing that Redux behaves as expected?

client/src/features/answers/answersSlice.test.js
```diff
@@ -0,0 +1,40 @@
...
+          answer: 5,
+        },
+      }
+    )
+  ).toEqual({
+    householdSize: 5,
+  });
+});
+
+test("answer update - boolean", () => {
```
egracer, line 25: Same as above -- update vs. add.

client/src/features/answers/answersSlice.js
mscharfstein: ack

client/src/features/answers/answersSlice.test.js
mscharfstein: when you import a variable without the { } it always imports the default export variable (in this case the reducer) from the file. the names don't have to be the same

egracer: Haha ooph that makes my head spin, good to know!

client/src/app/store.js
mscharfstein: changed name everywhere

PR #27, opened 2020-07-15
Make questions reducer with updateQuestions action

mscharfstein:
Works on #9 

Creates a "questions" slice - includes a questions object within the state (the keys are question identifiers like householdSize, and the values are the full question text), an action to update the questions, and a reducer to handle this action. The action will contain a payload with an object with the question keys and text. This object will be populated by calling the backend API (the question keys should be stored in the schema as well).  The question keys match the keys in the answers reducer (review https://github.com/ny/maslow/pull/26 first so that this makes more sense).

Note: An alternative option is to store the questions as a simple array without the key identifiers. However, then there is not a good way for the FE to store the mapping between question identifier and question text. The FE needs this mapping so that when a question is answered, it knows what identifier should be passed to the answers reducer.


client/src/features/questions/questionsSlice.js
```diff
@@ -0,0 +1,26 @@
+import { createSlice } from "@reduxjs/toolkit";
+
+export const questionsSlice = createSlice({
+  name: "questions",
+  initialState: {
+    // Will include keys representing the questions (which will match the keys
```
marcja, line 6: To be clear, the questions slice will only contain a map of questionKeys to metadata (which, currently, is just the question.type). The full text of the question will be pulled from i18next (via the `t` function), not from redux. As a hack, you can pull the full text from redux until i18next is fully configured and working, but that isn't the final design.

client/src/features/questions/questionsSlice.js
```diff
@@ -0,0 +1,26 @@
...
+    // in the answers slice of the state) and values with the full
+    // question text.
+    // Alternatively, we could just have an array of questions and not include
```
marcja, line 9: Check out my PR for updating the schema at #28. There is now a new `forms` array that contains the form structure (that is, the ordering of the questions in sections).

client/src/features/questions/questionsSlice.js
```diff
@@ -0,0 +1,26 @@
...
+    // the key, but then there would not be a good way for the FE to map from
+    // the question to the key that is used in the answers slice.
+  },
+  reducers: {
+    questionsUpdate: (state, action) => {
+      Object.keys(action.payload.questions).map(function (key) {
```
marcja, line 15: Maybe simplify to:
```javascript
Object.entries(action.payload.questions).map((key, val) => {
  state[key] = val;
});
```
Although, I'm curious... Wouldn't this work too:
```javascript
// note that this only is correct because redux toolkit uses immer
Object.assign(state, action.payload);
// if this were regular redux, it would have to be:
// Object.assign(state, {}, action.payload);
```

client/src/features/questions/questionsSlice.test.js
```diff
@@ -0,0 +1,27 @@
...
+});
+
+test("questions update", () => {
+  expect(
+    questions(
+      {},
+      {
+        type: questionsUpdate.type,
+        payload: {
+          questions: {
```
marcja, line 15: FYI, this does not match the current schema. See #28.

client/src/features/questions/questionsSlice.js
mscharfstein: good point. updated. but im not sure what you mean by hacking by pulling from redux... we would have to like hardcode into redux i guess? but sumer is working on i18n PRs so ill just let her get that set up

client/src/features/questions/questionsSlice.js
mscharfstein: ack

client/src/features/questions/questionsSlice.js
mscharfstein: the map(key, val) for some reason was val was representative of the index not the value. but the Object.assign works. i thought i should be able to just set state = action.payload (since redux toolkit uses immer so you can directly modify the state) but that wasn't working

client/src/features/questions/questionsSlice.test.js
mscharfstein: i updated, but Javascript doesn't have the idea of NUMBER or BOOLEAN so i just put it in as a string. is that what you were envisioning?

client/src/features/questions/questionsSlice.js
egracer: FWIW I think it would be:

```
Object.entries(action.payload.questions).map(([key, val]) => {
  state[key] = val;
});
```
since `Object.entries` returns an array.

client/src/features/questions/questionsSlice.js
marcja: See https://github.com/ny/maslow/blob/master/client/src/App.js#L25 for an example of the hack.

client/src/features/questions/questionsSlice.test.js
marcja: Yes. FWIW, you can fake an enum in JS like so:

```javascript
const QuestionType = {
  NUMBER: "NUMBER",
  BOOLEAN: "BOOLEAN"
};
Object.freeze(QuestionType);
```

Then you can use `question.type = QuestionType.NUMBER`. But that's all syntactic sugar.

client/src/features/questions/questionsSlice.js
mscharfstein: gotcha, just pulling the english title you mean. yeah - we can start with that (though not relevant for this PR really, because the current tests are sufficient)

client/src/features/questions/questionsSlice.test.js
mscharfstein: thanks! I think that adds a layer of complication that's not necessary but open to changing it to this in the future if we see it actually will help

Issue #25, opened 2020-07-14
Simplify schema for dynamic i18n content backend


itssumertime:
To have separate files for each language, we can configure spring.config.additional-location for a wildcard location in classpath (such as “classpath:locales/*/”) and create an application.yaml in each locale directory. The important wrinkle here is that Spring can only merge maps, not lists, so the current catalog schema will need to be modified to support this merging (for example, questions and services are currently lists). However, since questions are strictly ordered, we would need to add a level of indirection to the schema, creating an unordered map of question data plus a new array of question keys to define the question order. Note as well that this would mean that adding a new question or service would typically require editing eight files across eight directories, but it also means that a translator for a specific language would only have to look at one file.

marcja:
With #28, I've updated the schema to represent what we think is pretty close to the final schema (as viewed from the perspective of the client). I will do the work of breaking up `application.yaml` into separate language files in a follow-up PR. Before I get to that, though, I'm going to work on #24.

PR #28, opened 2020-07-15
Update schema per design

marcja:
Works on #25 

Per the design discussion yesterday, this updates the schema to reflect required changes. Specifically:

- **CHANGED**: The services array has been changed to a map. The name attribute has been removed, as it is now the key of the map.
- **CHANGED**: The questions array has been changed to a map. The name attribute has been removed, as it is now the key of the map.
- **ADDED**: A new sections map has been added. This contains the metadata/resources for sections.
- **ADDED**: A new forms array has been added. This contains the structure of the form (that is, the ordering of sections and the ordering of questions within their respective sections).
- **REMOVED**: The seeAlso field was removed, as it was not specified in the PRD.

Here's a screenshot of the new schema:

screencapture-localhost-8080-api-explore-v1-catalog-2020-07-15-08_51_43


server/src/main/java/gov/ny/its/hs/maslow/model/Catalog.java
```diff
@@ -22,36 +24,74 @@
...
+  private Map<String, Section> sections = new HashMap<>();
+
+  @NotEmpty
+  private Map<String, Question> questions = new HashMap<>();
+
+  @NotEmpty
+  private Map<String, Service> services = new HashMap<>();
+
+  /**
+   * Gets the structure definition of the form. Each form is comprised of one or more sections where
```
kebrahim, line 39: there's only one form right? "Each form" makes it sound like there are multiple...

server/src/main/java/gov/ny/its/hs/maslow/model/Catalog.java
```diff
@@ -22,36 +24,74 @@
...
+   * each section has one or more questions. The form structure is defined by reference, not by
```
kebrahim, line 40: can you add a little more information explaining how the java types correspond to the sections/questions? i.e. outer List -> list of sections, Map -> single section, Map key -> string id of section, Map value -> list of string ids of questions (i think that's right, but would like to confirm)

server/src/main/java/gov/ny/its/hs/maslow/model/Catalog.java
```diff
@@ -22,36 +24,74 @@
...
+
+  /**
+   * Sets the value of {@link #getForm()}
+   */
+  public void setForm(List<Map<String, List<String>>> form) {
+    this.form = form;
+  }
+
+  /**
+   * Gets the dictionary of sections that defines the metadata and resources that describes a
```
kebrahim, line 56: mind describing what the String key signifies?

server/src/main/java/gov/ny/its/hs/maslow/model/Catalog.java
```diff
@@ -22,36 +24,74 @@
...
+  /**
+   * Sets the value of {@link #getSections()}
+   */
+  public void setSections(Map<String, Section> sections) {
+    this.sections = sections;
+  }
 
   /**
-   * Gets the list of questions that we ask residents in order to recommend relevant services.
+   * Gets the dictionary of questions that we ask residents in order to recommend relevant services.
```
kebrahim, line 71: similarly, please describe what the String key signifies

server/src/main/java/gov/ny/its/hs/maslow/model/Catalog.java
```diff
@@ -22,36 +24,74 @@
...
    * Sets the value for {@link #getQuestions()}
    */
-  public void setQuestions(List<Question> questions) {
+  public void setQuestions(Map<String, Question> questions) {
     this.questions = questions;
   }
 
   /**
-   * Gets the list of services that we can recommend to residents.
+   * Gets the dictionary of all services that we can recommend to residents.
```
kebrahim, line 85: similarly, please describe what the String key signifies

server/src/main/java/gov/ny/its/hs/maslow/model/Question.java
```diff
@@ -2,40 +2,19 @@
...
-import org.springframework.data.annotation.Id;
+import javax.validation.constraints.NotEmpty;
+import javax.validation.constraints.NotNull;
 
 /**
  * Defines a question that is asked the resident. The question also defines a variable that can be
  * referenced by a formula in each service for matching service recommendations to residents.
  */
 public class Question {
-  @Id
```
kebrahim, line 13: why doesn't the Question have an Id anymore?

server/src/main/java/gov/ny/its/hs/maslow/model/ServiceResources.java
```diff
@@ -49,32 +49,32 @@ public void setCategory(String category) {
...
    */
-  public void setAliases(String aliases) {
-    this.aliases = aliases;
+  public void setSummary(String summary) {
+    this.summary = summary;
   }
 
   /**
-   * Gets a short description of the service.
+   * Gets a comma-separated list of aliases that this service may go by. For example, "Supplemental
```
kebrahim, line 66: sorry, just noticed this. why do comma-separated, considering that will limit what characters can be included in the alias name? would it make more sense to have an array of Strings?

server/src/main/resources/application.yaml
```diff
@@ -1,174 +1,186 @@
...
-      type: BOOLEAN
-      resources:
-        en:
-          text: Are you or a member of your household currently a student?
+    questions:
+      HOUSEHOLD_SIZE:
+        type: NUMBER
+        resources:
+          en:
+            text: How many people live at your current address (including you)?
```
kebrahim, line 54: this can be a TODO, but i just wanted to note this doesn't match the mocks or the 'service selection criteria by service' doc (https://docs.google.com/document/d/19cDS3dL9ZtoSPXFHuJy9haW3Pdr9aTwwwiVpAJxNJh0/edit#heading=h.s99ou2oodp23). honestly, i'm not sure which of these should be the source of truth

server/src/main/java/gov/ny/its/hs/maslow/model/Catalog.java
marcja: Yep, only one form. Will fix.

server/src/main/java/gov/ny/its/hs/maslow/model/Question.java
marcja: Because the question is contained in a map and the id (key) is outside the question.

server/src/main/java/gov/ny/its/hs/maslow/model/ServiceResources.java
marcja: It's not parsed, just displayed. I had it as an array of strings in an earlier version, but I didn't think it added anything other than complexity.

server/src/main/resources/application.yaml
marcja: All of the UI text needs to be reviewed. I made a comment to Malika and in the UX review that Code For America did a bunch of UXR and discovered that "household" is a very confusing term. They tested a bunch of variations and found the "at your current address" was more clearly understood. Anyway, all of this data is provisional anyway until we get it reviewed.

server/src/main/java/gov/ny/its/hs/maslow/model/Catalog.java
marcja: fixed

server/src/main/java/gov/ny/its/hs/maslow/model/Catalog.java
marcja: fixed

server/src/main/java/gov/ny/its/hs/maslow/model/Catalog.java
marcja: fixed

server/src/main/java/gov/ny/its/hs/maslow/model/Catalog.java
marcja: fixed

server/src/main/java/gov/ny/its/hs/maslow/model/Catalog.java
marcja: fixed

server/src/main/resources/application.yaml
```diff
@@ -1,174 +1,186 @@
...
-        es:
-          text: ¿Está trabajando actualmente?
+    sections:
+      SECTION_HOUSEHOLD:
+        resources:
+          en:
+            title: Household size
+          es:
+            title: Tamaño del hogar
+          ru:
```
itssumertime, line 26: Didn't we decide to do the languages in separate files?

server/src/main/java/gov/ny/its/hs/maslow/model/Question.java
itssumertime: Are you saying we use the keys of the Question map as the id for it?

marcja: Correct.

server/src/main/resources/application.yaml
marcja: That work is still planned, just not part of this PR. In terms of priorities, this schema update unblocks the other engineers. I'm going to work on the dynamic content backend first, then come back to splitting the language files as a cleanup task.

kebrahim: similar question. is it just a TODO to move these strings to separate yaml files?

Issue #10, opened 2020-07-14
Create the summary page component


egracer:
Depends on #9.

Build and test a React component that allows the user to review the answers they provided to all questions previously displayed.

1. Write a [selector](https://react-redux.js.org/api/hooks#useselector-examples) to retrieve the answers to all questions from Redux.
2. Build a component to display the questions & their corresponding answers as determined in step 2. This does not need to be complicated (as this screen may be discarded in a future version of the UXD).

Requirements:

- The order of questions & answers in the summary should match the ordering that the questions were displayed.
- The summary page should have a back button which returns the user to the last page of questions.

egracer:
Tasks remaining:

- [ ] Connect to Redux.
- [ ] The final question page links to this component.
- [ ] The `Back` button links to the final question page.

PR #29, opened 2020-07-15
Simple summary page

egracer:
Works on #10.

Wrote a simple UI to display question responses based on the schema from #28. Not writing tests since there is a strong possibility that this page will only be for development/demo purposes.

Screenshot:

qO9ZiZvT12y


client/src/components/SummaryPage.jsx
```diff
@@ -0,0 +1,33 @@
+import React, { useContext } from "react";
+
+/**
+ * PLACEHOLDER data being used until the redux store is connected.
+ */
+const data_ = 
+  {
+    houseHoldSize: 5,
```
marcja, line 8: Please see #28 for what the data will actually look like. For example, this will be:
```
const data_ = {
  "HOUSEHOLD_SIZE": 5,
  "IS_EMPLOYED: true
};

File: client/src/components/SummaryPage.jsx [hunk 2]
```diff
@@ -0,0 +1,33 @@
...
+  const answers = useContext(AnswersContext);
+
+  return (
+    <div>
+      <h1>Summary</h1>
+      {Object.entries(answers).map(([questionKey, response]) => (
+        <p key={questionKey}>{questionKey}: {String(response)}</p>
+      ))}
+      {/* TODO: currently a noop, needs to be hooked up to the last of the question group pages */}
+      <button>Back</button>
```
marcja, line 28: More of a UX question, but is a custom back button necessary in a web app? The browser already has a back button.

client/src/components/SummaryPage.jsx
```diff
@@ -0,0 +1,33 @@
...
+    </div>
+  );
+}
+
+export default SummaryPage;
```
marcja, line 33: Add newline at end of file.

client/src/components/SummaryPage.jsx
egracer: Fair point, let's ask Malika. I would imagine that a back button in the web app would be easier for less experienced users or mobile users when browser controls are hidden.

client/src/components/SummaryPage.jsx
egracer: Done

client/src/components/SummaryPage.jsx
egracer: Done

client/src/components/SummaryPage.jsx
mscharfstein: if a user presses the browser back button, does our redux state remain? i think yes, but just want to make sure. also i do think its helpful to have back and forward arrows when showing the questions so we might want to stay consistent and have a back arrow on this page too

client/src/components/SummaryPage.jsx
```diff
@@ -0,0 +1,33 @@
...
+/**
+ * PLACEHOLDER data being used until the redux store is connected.
+ */
+const data_ = 
+  {
+    "HOUSEHOLD_SIZE": 5,
+    "IS_EMPLOYED": true
+  }
+
+const AnswersContext = React.createContext(data_);
```
mscharfstein, line 12: what does this do? can you add a comment? also, let's call everything Responses because I changed the Reducer to do that

mscharfstein: oh i see, this is just temporary along with the data above. nvm

client/src/components/SummaryPage.jsx
```diff
@@ -0,0 +1,33 @@
...
+
+/**
+ * Displays the question responses submitted by the user.
+ */
+function SummaryPage() {
+  /* TODO: This line replaced by Redux Toolkit's useSelector() */
+  const answers = useContext(AnswersContext);
```
mscharfstein, line 19: s/answers/responses

client/src/components/SummaryPage.jsx
egracer: Spoke with Malika. Without an explicit button or link, a user would have to infer that navigating to a previous step required using the back button. For less technically experienced or confident users, that could present a usability barrier.

client/src/components/SummaryPage.jsx
egracer: Correct, just temporary.

Issue #13, opened 2020-07-14
Render the QuestionForm


tariqajyusuf:
Implement functionality in React to retrieve catalog from Redux, parse the groups & questions.

Storing the state is being tracked in another issue and is not in scope here. This issue is just tracking the creation of a React-based representation of the catalog objects.

Handles displaying the relevant page, and navigation forward & backward

- QuestionForm contains a single Section component
  - passes in section_key, list of question_keys into Section component
  - Section reads & renders section info (title) from i18next using SectionKey
  - Section reads question type from Redux using questionKeys, in order to know which question component to render
- QuestionForm contains forward/back button, "Step # of #" label

kebrahim:
- we're going to store all data from the API in Redux. App.js will take care of that. it already has the call to the API; it just needs to be updated to call the appropriate slices to store the data for the form, the sections & the questions [kebrahim@ will add this as part of #12 ]
- in order to do that, new slices for the form & the sections need to be created [mscharfstein@ will add this as part of #9 ]
- we'll have a Form component that will read in the form data from the formSlice, which will include the ordering of the sections, the ordering of the questions w/in those sections & the currentSection. the Form component will render all of the Section components (passing in the SECTION_ID to each component], displaying the current Section component & hiding the rest [kebrahim@ will work on this as part of #13 ]
- the Section component will be passed in the SECTION_ID from the Form component, it will look up the ordering of the questions from the formSlice, the section data from the sectionSlice & then render the section information & the individual Question components, passing in the QUESTION_ID to each component [yusuft@ will work on this as part of #14 ]
- the Question component will look up the question data from the questionSlice & then based on the type, will render the BooleanQuestion or IntegerQuestion components [egracer@ & mscharfstein@ will work on this as part of #15 & #16 ]

kebrahim:
Navigation design:
- QuestionForm contains a single Section component
- QuestionForm passes in section_key, list of question_keys into Section component
- Section reads & renders section info (title) from i18next using SectionKey
- Section reads question type from Redux using questionKeys, in order to know which question component to render
- QuestionForm contains forward/back button, "Step # of #" label

kebrahim:
Tests should include (from #72 ):
- submit button works properly (mock "history")
- check that nothing displays if sections/questions are empty (useSelector mocked to mock return values for form, sections, questions)
- check that the right number of sections are displayed
- check that the right number of questions per section are displayed (this will move to the Sections component testing once Sections is built)

PR #33, opened 2020-07-15
Add simple router to top level of application

kebrahim:
Works on #13 

I just wanted to make sure we all agreed that a Router makes sense, which will route the user to the Questionnaire component (coming in a future PR).


client/src/Routes.jsx
```diff
@@ -0,0 +1,19 @@
...
+import {Route, Switch} from 'react-router-dom';
+import App from './App';
+
+/**
+ * @return {string} Routes the supported paths to components
+ */
+function Routes() {
+  // TODO(kebrahim): Switch to Questionnaire component
+  return (
+    <div>
```
marcja, line 11: Is there a more semantic HTML element than `<div>`? Maybe `<section>`?

kebrahim: done

Issue #14, opened 2020-07-14
Render section


tariqajyusuf:
The rendering of question groups should show:

- [x] The title of the group
- [x] A list of the boolean and integer questions in this group
- [x] unit testing
- [x] translation

Interaction with QuestionForm:
- QuestionForm passes in section_key, list of question_keys into Section component
- Section reads & renders section info (title) from i18next using SectionKey
- Section reads question type from Redux using questionKeys, in order to know which question component to render

Fields moved to QuestionForm, based on new navigation design:
- How many total groups there are and how far we are (e.g. "Step # of #")
- A back link
- A continue button to go to the next group

tariqajyusuf:
Initial skeleton is committed in #36, working on integrating reading from Redux.

kebrahim:
updated list of section tasks, based on new navigation design - https://github.com/ny/maslow/issues/13#issuecomment-661327339

PR #36, opened 2020-07-15
feat: Add Section component for rendering part of questionnaire

tariqajyusuf:
First attempt at a question section component for #14


client/src/components/Section.jsx
```diff
@@ -0,0 +1,50 @@
+import React from "react";
+
+class Section extends React.Component {
```
kebrahim, line 3: we haven't been extending React.Component in other files. maybe michelle knows better about when & why that's necessary? i only point this out because it's inconsistent with the rest of the files, but i'm not sure what the best practice should be

client/src/components/Section.jsx
```diff
@@ -0,0 +1,50 @@
...
+            // TODO: Pull invalid statement from question object.
+            return <div>Question {questionId} is invalid.</div>;
+          }
+          return "";
+        })}
+      </div>
+    );
+  }
+
+  render() {
```
kebrahim, line 25: the sectionId should be passed into this class somehow. then it should be able to read the relevant information from Redux

client/src/components/Section.jsx
```diff
@@ -0,0 +1,50 @@
+import React from "react";
+
+class Section extends React.Component {
+  generateQuestions(questionIds) {
```
kebrahim, line 4: add method-level documentation here & the rest of the methods below

client/src/components/Section.jsx
```diff
@@ -0,0 +1,50 @@
+import React from "react";
```
kebrahim, line 1: if you execute `yarn lint --fix` in the `client/` directory, it should clean up quite a few of the linting errors, e.g. use of double-quotes

client/src/components/Section.jsx
```diff
@@ -0,0 +1,50 @@
...
+      <h2>This is where we'd initialize a question component for ID {id}.</h2>
+    ));
+  }
+
+  renderValidation(questionIds, validityState) {
+    return (
+      <div class="validation-error">
+        {questionIds.map((questionId) => {
+          if (!validityState.get(questionId)) {
+            // TODO: Pull invalid statement from question object.
```
kebrahim, line 16: wouldn't the question component handle this?

client/src/components/Section.jsx
kebrahim, line 3: add class-level documentation

kebrahim, line 3: please also add a test class

marcja: Extending React.Component is the "old" style. The "new" style is functional and uses hooks. We should definitely all be using the "new" style.

client/src/components/Section.jsx
```diff
@@ -0,0 +1,50 @@
...
+    const validityState = new Map([
+      ["q1", false],
+      ["a2", true],
+    ]);
+
+    var questionElements = this.generateQuestions(questionIds);
+    var validationErrors = this.renderValidation(questionIds, validityState);
+
+    return (
+      <div>
```
marcja, line 37: I'm assuming that each <Section> actually models a <fieldset> (just as <Form> models a <form>).

client/src/components/Section.jsx
```diff
@@ -0,0 +1,80 @@
...
+   *  status.
+   * @return {string} Rendered validation text or nothing if the state is
+   *  valid.
+   */
+  function renderValidation(questionIds, validityState) {
+    return (
+      <div className="validation-error">
+        {questionIds.map((questionId) => {
+          if (!validityState.get(questionId)) {
+            // TODO: Pull invalid statement from question object.
```
kebrahim, line 43: probably a question for the group, but should we be pulling the invalid statement from Redux, instead?

client/src/components/Section.jsx
```diff
@@ -0,0 +1,80 @@
...
+  const questionIds = ["q1", "a2"];
+  const validityState = new Map([
+    ["q1", false],
+    ["a2", true],
+  ]);
+
+  const questionElements = generateQuestions(questionIds);
+  const validationErrors = renderValidation(questionIds, validityState);
+
+  return (
```
kebrahim, line 62: add TODO to display section name

client/src/components/Section.jsx
```diff
@@ -0,0 +1,80 @@
...
+    <fieldset>
+      <h1 data-testid="section-header">
+        Step {props.currentStep} of {props.totalSteps}
```
kebrahim, line 65: i think you can get both of these values from Redux & not require Form to pass it in as a prop

client/src/components/Section.jsx
```diff
@@ -0,0 +1,80 @@
...
+ */
+function Section(props) {
+  /**
+   * Calls rendering of each of the questions in this section.
+   *
+   * @param {Array<String>} questionIds list of question IDs, will eventually
+   *  be removed in favor of reading from Redux.
+   * @return {string} Rendered questions for this section.
+   */
+  function generateQuestions(questionIds) {
```
marcja, line 18: s/questionIds/questionKeys/g (for consistency with other components)

client/src/components/Section.jsx
```diff
@@ -0,0 +1,80 @@
...
+    // TODO: Render the actual questions.
+    return questionIds.map((id) => (
```
marcja, line 20: s/id/questionKey/g

client/src/components/Section.jsx
marcja, line 18: s/generateQuestions/renderQuestions/g?

client/src/components/Section.jsx
```diff
@@ -0,0 +1,80 @@
...
+    ["q1", false],
+    ["a2", true],
+  ]);
+
+  const questionElements = generateQuestions(questionIds);
+  const validationErrors = renderValidation(questionIds, validityState);
+
+  return (
+    <fieldset>
+      <h1 data-testid="section-header">
```
marcja, line 64: Strictly speaking, this should be a `<legend>` instead of an `<h1>`, but I'm not exactly sure how semantic-ui renders that.

client/src/components/Section.jsx
```diff
@@ -0,0 +1,80 @@
...
+   * @param {Map<String, Boolean>} validityState list of validation states,
+   *  will eventually be removed in favor of reading from each component's
+   *  status.
+   * @return {string} Rendered validation text or nothing if the state is
+   *  valid.
+   */
+  function renderValidation(questionIds, validityState) {
+    return (
+      <div className="validation-error">
+        {questionIds.map((questionId) => {
```
marcja, line 41: Given that each question will also render its validity state, this should probably just provide a brief summary. Do the UX mocks show this?

client/src/components/Section.jsx
```diff
@@ -0,0 +1,80 @@
...
+  const validationErrors = renderValidation(questionIds, validityState);
+
+  return (
+    <fieldset>
+      <h1 data-testid="section-header">
+        Step {props.currentStep} of {props.totalSteps}
+      </h1>
+      {validationErrors}
+      {questionElements}
+      <button>Continue</button>
```
marcja, line 69: Add TODO for progressing between sections

client/src/components/Section.jsx
tariqajyusuf: No I don't think, so. I'll just put a generic message then.

client/src/components/Section.jsx
tariqajyusuf: Closing in favor of marc's comment above to use a generic phrase.

client/src/components/Section.jsx
tariqajyusuf: Yes, I'll fix this in the next PR when I pull from Redux.

client/public/locales/ko/translation.json
```diff
@@ -7,5 +7,4 @@
   "summary": {
     "title": "요약"
   }
-
```
kebrahim, line 10: revert

client/src/components/Section.test.js
```diff
@@ -0,0 +1,12 @@
+import React from "react";
+import { render } from "@testing-library/react";
+import Section from "./Section.jsx";
+
+test("header renders correctly", () => {
```
kebrahim, line 5: i changed my PR to use the following format to help w/ readability:
```
describe("...", () => {
  it("...", () => {
    ...
  });
});
```
please do the same here

client/src/components/Section.jsx
```diff
@@ -0,0 +1,88 @@
...
+   * Calls rendering of each of the questions in this section.
+   *
+   * @param {Array<String>} questionKeys list of question IDs, will eventually
+   *  be removed in favor of reading from Redux.
+   * @return {string} Rendered questions for this section.
+   */
+  function renderQuestions(questionKeys) {
+    // TODO: Render the actual questions.
+    return questionKeys.map((questionKey) => (
+      <h2 key={questionKey}>
```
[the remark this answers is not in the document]
tariqajyusuf: Agreed, I'll bring that in as I add Redux stuff in the next PR.

client/public/locales/ko/translation.json
tariqajyusuf: This was done by the linter. So I'm still planning to commit this.

client/src/components/Section.jsx
marcja: FYI, no longer am convinced that these particular two fields should be pulled from redux

Issue #12, opened 2020-07-14
Add reading from / writing to Redux from the question pages


egracer:
Depends on #9.

Implement [selectors](https://react-redux.js.org/api/hooks#useselector-examples) & [action functions](https://redux-toolkit.js.org/tutorials/basic-tutorial#introducing-createslice) generated by `createSlice` within the question pages. The purpose of implementing this feature is to support:

1. Showing the existing answers to questions as the user navigates back and forth between question pages, and 
2. To display the answers to all questions on the summary page at the end of the form.
3. Load the questions

This task is complete when questions are saved to `questions` slice, answers are saved to the `answers` slice when a user completes a question, and when the answers are successfully loaded back into the component after navigating away from and back into the corresponding question component.

Note: testing that the Redux actions properly mutate and save state falls under the purview of #9.

kebrahim:
done

PR #37, opened 2020-07-16
Store API question data in Redux, query on new Form component

kebrahim:
Works on #12 #13 

In App.js, the question API data is now stored in Redux instead of displayed immediately. Also, relays rendering to a new Form component, which will be in charge of reading form data from Redux and rendering the entire questionnaire form. For now, it just verifies that the question data can be read from Redux & rendered.

vHfQVQinjCA


client/src/App.test.js
```diff
@@ -4,12 +4,13 @@ import {Provider} from 'react-redux';
 import store from './app/store';
 import App from './App';
 
-test('renders learn react link', () => {
+// TOOD(kebrahim): Add tests confirming data is stored in Redux
+test('renders questionnaire form', () => {
```
marcja, line 8: just 'renders form' is fine

client/src/Routes.test.js
```diff
@@ -6,13 +6,13 @@ import Routes from './Routes';
 import {Provider} from 'react-redux';
 import store from './app/store';
 
-test('/ routes to App', async () => {
+test('/ routes to Questionnaire', async () => {
```
marcja, line 9: s/Questionnaire/Form/g?

client/src/App.test.js
kebrahim: done

client/src/Routes.test.js
kebrahim: done

PR #39, opened 2020-07-16
add endpoint to retrieve localized catalog text

marcja:
# Features

This adds an endpoint that implements the GET methods:

- /api/explore/v1/locales/{language}/catalog.json
- /api/explore/v1/locales/{language}/sections.json
- /api/explore/v1/locales/{language}/questions.json
- /api/explore/v1/locales/{language}/services.json

where language is the locale identifier, such as "en", "es", "ko", "ru", "zh", "bn", or "ht".

Each method returns JSON that matches the [i18next JSON v3](https://www.i18next.com/misc/json-format#i-18-next-json-v3) format for a particular language. JSON in this format can be imported directly into i18next via [addResourceBundle](https://www.i18next.com/overview/api#addresourcebundle). 

In practice, [i18next-http-backend](https://github.com/i18next/i18next-http-backend) will handle loading these resources and adding them to i18next. The URLs for these endpoints where designed to match as closely as possible the default path used by i18next-http-backend ("/locales/{{lng}}/{{ns}}.json"), where `lng` is the language and `ns` is the i18next namespace. Here, "catalog", "sections", "questions", and "services" are the i18next namespaces.

The catalog namespace combines the resources from the sections, questions, and services namespaces into a single namespace.

# Testing

There is remarkably little code here, and one would think it would be easy to test. If in fact this were plain Java and not Spring Boot, it probably would be. But there is a lot of Spring Boot magic going on, and it's taking a long time to figure out the right incantations to get the controller tests to have all of their autowired dependencies and configuration provided and/or mocked. I'm continuing to work on it, but I didn't want to block others on it.


server/src/main/java/gov/ny/its/hs/maslow/model/ResourcesProvider.java
```diff
@@ -0,0 +1,16 @@
...
+ * Defines an generic interface for classes that return a resources objects (that is, a map of
+ * locale to translated strings).
+ * @param <T> a resources class (such as {@link SectionResources}, {@link QuestionResources}, or
+ * {@link ServicesResources})
+ */
+public interface ResourcesProvider<T> {
+  /**
+   * Gets the resources (that is, a map of locale to translated strings)
+   */
+  public Map<String, T> getResources();
```
itssumertime, line 15: Would this be used by the classes implementing it to be like "question_type.getResources()" and "section.getResources()"? And since it's a map we would index into it based on the language to get the list of all the strings? Is the key a language or a phrase?

server/src/main/java/gov/ny/its/hs/maslow/model/ResourcesProvider.java
```diff
@@ -0,0 +1,16 @@
+package gov.ny.its.hs.maslow.model;
+
+import java.util.Map;
+
+/**
+ * Defines an generic interface for classes that return a resources objects (that is, a map of
```
itssumertime, line 6: s/an/a

kebrahim, line 6: s/objects/object

server/src/main/java/gov/ny/its/hs/maslow/model/ResourcesProvider.java
marcja: This is implemented by the `Section`, `Question`, and `Services` classes that each have a matching getter for their respective resources (`SectionResources`, `QuestionResources`, and `ServiceResources`). I created this interface simply to make the code for `LocalesController` cleaner (see that class for usage). The resulting map is in the form:

screencapture-localhost-8080-api-explore-v1-locales-en-catalog-json-2020-07-16-15_06_16

server/src/main/java/gov/ny/its/hs/maslow/model/ResourcesProvider.java
marcja: fixed

marcja: fixed

Issue #22, opened 2020-07-14
Integrate useTranslation hook and "t" function for translations


itssumertime:
Use the i18next "t" function and "useTranslation" hook to fetch the appropriate language for each piece of content.

Issue #21, opened 2020-07-14
Create i18n static content backend


itssumertime:
Make directories for each locale with accompanying resources.json file and use [ i18next-http-backend](https://github.com/i18next/i18next-http-backend) to load in the resources.

- [ ] mkdir -p client/public/locales/[en,es,ru,zh,ko,bn,ht]/
- [ ] add resources.json to each directory
- [ ] configure i18next.js to use i18next-http-backend

PR #41, opened 2020-07-16
Add i18n static content backend and useTranslation example

itssumertime:
Addresses issue #21 and #22. 

Adds an example of the "t" function with "useTranslation".
Adds example phrases into "en" (English) and "es" (Spanish) translation.json files (other languages will be added when more content exists).


itssumertime:
I had to merge quite a few things from master so please let me know if something looks off!

client/src/App.js
```diff
@@ -12,6 +13,7 @@ import './App.css';
  */
 function App() {
   const dispatch = useDispatch();
+  const [t, i18n] = useTranslation();
```
marcja, line 16: Should be:
```javascript
const {t, i18next} = useTranslation();
```

Be sure to run `yarn lint`

client/src/App.js
```diff
@@ -26,14 +28,12 @@ function App() {
       dispatch(sectionsUpdate(data));
       dispatch(questionsUpdate(data));
       dispatch(servicesUpdate(data));
+      
+      {t("TRANSLATION_EXAMPLE")};
```
marcja, line 32: Not sure that this code makes sense here inside of useEffect, even as sample code...

client/src/i18n.js
```diff
@@ -1,5 +1,12 @@
+import { initReactI18next } from "react-i18next";
+import LanguageDetector from "i18next-browser-languagedetector";
+import translationEN from "./locales/en/translation.json";
+import translationES from "./locales/es/translation.json";
+=======
```
marcja, line 5: What is this? Merge residue?

client/src/i18n.js
```diff
@@ -1,5 +1,12 @@
...
 import i18n from 'i18next';
+import i18next from 'i18next';
```
marcja, line 7: You don't need this and the previous line. Delete this line and reference `i18n` everywhere instead.

client/src/i18n.js
```diff
@@ -1,5 +1,12 @@
...
 import {initReactI18next} from 'react-i18next';
+import HttpApi from 'i18next-http-backend';
 import LanguageDetector from 'i18next-browser-languagedetector';
 import translationEN from './locales/en/translation.json';
 import translationES from './locales/es/translation.json';
```
marcja, line 12: You don't want to import these two translations files here. The http-backend is supposed to load this files.

client/src/i18n.js
```diff
@@ -14,12 +21,49 @@ const resources = {
   },
 };
```
marcja, line 22: You don't want this either. You're mixing two different methods. Either you want to statically import the JSON files (are you're doing here) or use the http-backend, but not both.

client/src/i18n.js
```diff
@@ -14,12 +21,49 @@ const resources = {
...
+    credentials: "same-origin",
+    cache: "default",
+  },
+
+  reloadInterval: false, // can be used to reload resources in a specific interval (useful in server environments)
+};
+
+i18next.use(HttpApi).init({
+  backend: i18nextOptions,
+});
```
marcja, line 56: You can delete lines 24-56. Most of this isn't needed.

client/src/i18n.js
```diff
@@ -14,12 +21,49 @@ const resources = {
...
+
 i18n
-    .use(LanguageDetector)
-    .use(initReactI18next) // passes i18n down to react-i18next
-    .init({
-      resources,
-      fallbackLng: 'en',
-    });
+  .use(LanguageDetector)
```
marcja, line 59: add the following above this line:
```javascript
  .use(HttpApi)
```
We don't require any custom configuration of http-backend at the moment.

client/src/i18n.js
```diff
@@ -14,12 +21,49 @@ const resources = {
...
+  .use(initReactI18next) // passes i18n down to react-i18next
+  .init({
+    react: {
+      useSuspense: false,
+    },
```
marcja, line 64: Why did you disable useSuspense? With this setting, I believe it's possible to flash untranslated content...

client/src/locales/en/translation.json
```diff
@@ -1,3 +1,4 @@
 {
-  "HOUSEHOLD_SIZE": "Household size"
+  "HOUSEHOLD_SIZE": "Household size",
```
marcja, line 2: Remove this line. I find it confusing to reference server-provided content from this file.

client/src/locales/es/translation.json
```diff
@@ -1,3 +1,4 @@
 {
-  "HOUSEHOLD_SIZE": "Tamaño del hogar"
+  "HOUSEHOLD_SIZE": "Tamaño del hogar",
```
marcja, line 2: Remove this line. I find it confusing to reference server-provided content from this file.

client/src/i18n.js
itssumertime: Hmm I don't see it in my branch but yeah looks like it. Thought I resolved them all.

client/src/i18n.js
itssumertime: I don't understand. I thought you said you want the http options, but are you saying that it will just use default options then?

client/src/i18n.js
itssumertime: I needed this for when I clicked translate to Russian but I didn't have a Russian file available, so it needed to be set as backup. I can take it out for now until I add the dropdown but I think it'll be necessary again.

client/src/App.js
itssumertime: Actually since I took out the dropdown which uses it, the "i18next" in this line is unused so I'm removing it and can add it back in the next PR.

client/src/App.js
itssumertime: Oops didn't notice it was inside. Moved it below the function.

client/src/i18n.js
itssumertime: Cleaned it.

client/src/i18n.js
itssumertime: Also that's pretty much what I've already pushed earlier although I guess it should leverage the HttpApi now.

client/src/App.js
```diff
@@ -1,24 +1,25 @@
-import React, {useEffect} from 'react';
-import {useDispatch} from 'react-redux';
-import {formUpdate} from './features/form/formSlice';
-import {sectionsUpdate} from './features/sections/sectionsSlice';
-import {servicesUpdate} from './features/services/servicesSlice';
-import {questionsUpdate} from './features/questions/questionsSlice';
-import Form from './components/Form.jsx';
-import './App.css';
+import { useEffect } from "react";
```
mscharfstein, line 1: i think this loses the linting. try running yarn lint --fix

itssumertime: Yeah idk why it does that every time I hit save but I think it's fixed now.

client/src/App.js
```diff
@@ -1,24 +1,26 @@
...
+import { useTranslation } from "react-i18next";
+import Form from "./components/Form.jsx";
+import "./App.css";
 
 /**
  * @return {string} The initial page of the application
  */
 function App() {
   const dispatch = useDispatch();
+  const [t] = useTranslation();
```
marcja, line 16: You need to use object syntax, not array syntax:

```javascript
const {t} = useTranslation();
```

See https://react.i18next.com/latest/usetranslation-hook. 
```

File: client/src/App.js [hunk 14]
```diff
@@ -31,9 +33,11 @@ function App() {
     fetchData();
   });
 
-  return (
-    <Form />
-  );
+  {
+    t("TRANSLATION_EXAMPLE");
+  }
```
marcja, line 38: Can you put a TODO/HACK comment on this? This is still clearly out of place.

itssumertime: Yes.

client/src/i18n.js
```diff
@@ -1,20 +1,9 @@
-import i18n from 'i18next';
 import {initReactI18next} from 'react-i18next';
+import HttpApi from 'i18next-http-backend';
 import LanguageDetector from 'i18next-browser-languagedetector';
-import translationEN from './locales/en/translation.json';
-import translationES from './locales/es/translation.json';
-
-// The translations.
-const resources = {
```
mscharfstein, line 8: how do you use resources on line 10 below if you aren't specifying it here?

itssumertime: Removed it.

client/src/i18n.js
itssumertime: I do end up needing this so I added it back again. The error without it is:
"Error: Form suspended while rendering, but no fallback UI was specified.
Add a <Suspense fallback=...> component higher in the tree to provide a loading indicator or placeholder to display."

PR #43, opened 2020-07-16
Create component for handling questions of type 'NUMBER'

### LINKED ISSUE #16: Render integer questions
Questions involving integers (e.g. "What is your annual income?") should have:

- [x] The question title
- [x] A question hint
- [x] A textbox with integer validation
- [x] Validation and error text logic to require non-negative integers (>=0)
- [x] Use HTML5 validation to use numeric keypad
- [x] Testing

egracer:
Fixes #16.

Features of the component:

- Only accepts positive integers.
- Pulls question & hint text from Redux (temporarily until i18n is adopted)
- Stores & pulls the user's response to the question in Redux.
- Uses HTML5 form attributes to enforce that only a number can be entered. Uses manual JS validation to ensure that the number entered is a positive integer.
- Styled as a [`Form`](https://react.semantic-ui.com/collections/form/) using [`semantic-ui-react`](https://react.semantic-ui.com/). Incorporates color changes and a prompt when the value entered in the input field is invalid.

Screenshots:

On load
Valid input
Invalid input


client/public/index.html
```diff
@@ -3,6 +3,7 @@
   <head>
     <meta charset="utf-8" />
     <link rel="icon" href="%PUBLIC_URL%/favicon.ico" />
+    <link rel="stylesheet" href="//cdn.jsdelivr.net/npm/semantic-ui@2.4.2/dist/semantic.min.css" />
```
egracer, line 6: I tried to add this in as a CSS package, but was not able to get the CSS to load from `node_modules` for some reason.

client/src/components/Form.jsx
```diff
@@ -39,6 +40,7 @@ function Form() {
     <div>
       <h1>Form</h1>
       {renderForm(form)}
+      <IntegerQuestion questionKey='HOUSEHOLD_INCOME' />
```
marcja, line 43: Any reason this isn't `NumberQuestion` instead of `IntegerQuestion`? The type is NUMBER, after all, not INTEGER.

client/src/components/IntegerQuestion.jsx
```diff
@@ -0,0 +1,72 @@
...
+ * positive integer response.
+ */
+function IntegerQuestion(props) {
+  const {questionKey} = props;
+  const a11yId = `integer-question-${questionKey}`;
+
+  const dispatch = useDispatch();
+  const responseValue = useSelector(selectResponses)[questionKey];
+  const questionData = useSelector(selectQuestions)[questionKey];
+  const [validationState, setValidationState] = useState(ValidationState.VALID);
```
marcja, line 30: Add a TODO here. I assume that this state is to be moved into Redux...

client/src/components/IntegerQuestion.jsx
```diff
@@ -0,0 +1,72 @@
...
+
+  // TODO: replace with i18next
+  const questionStrings = questionData.resources.en;
+
+  const handleInput = (event) => {
+    const newValue = event.target.value;
+    if (!newValue || newValue.includes('-') || newValue.includes('.')) {
```
marcja, line 37: Prefer regular expression:
```javascript
const NUMBER_PATTERN = /\d+/;
...
if (!NUMBER_PATTERN.test(newValue) { ... }
```

client/src/components/IntegerQuestion.jsx
```diff
@@ -0,0 +1,72 @@
...
+  };
+
+  return (
+    <Form size='large'>
+      {/* TODO: `Form` should be moved to the section component */}
+      <Form.Field required error={!!validationState}>
+        {/* eslint-disable-next-line jsx-a11y/label-has-associated-control */}
+        <label htmlFor={a11yId}>{questionStrings.text}</label>
+        {!!questionStrings.hint &&
+          <p style={{'font-size': '1rem'}}>{questionStrings.hint}</p>}
```
egracer, line 56: Not sure what the right place is to put this styling, but I did confirm that Semantic UI does not have built-in support for predefined text sizes. We definitely want to adopt the same hint text size across both number & boolean question components.

client/src/components/IntegerQuestion.jsx
```diff
@@ -0,0 +1,72 @@
...
+    if (!newValue || newValue.includes('-') || newValue.includes('.')) {
+      setValidationState(ValidationState.INVALID);
+    } else {
+      setValidationState(ValidationState.VALID);
+    }
+
+    // Note: we are always saving the value that the user enters to prevent
+    // inconsistent state between the Redux store and the value displayed in the
+    // input box.
+    dispatch(responsesUpdate({questionKey: questionKey, response: newValue}));
```
marcja, line 46: I expected that the data would be:
```javascript
{ "HOUSEHOLD_SIZE": 4 }
```
not:
```javascript
{ questionKey: "HOUSEHOLD_SIZE", response: 4 }
```
What was the reasoning for the latter?

client/src/components/IntegerQuestion.jsx
```diff
@@ -0,0 +1,72 @@
...
+  };
+
+  return (
+    <Form size='large'>
```
marcja, line 50: Is this temporary? I'm assuming that the `<form>` element should be defined at a higher level (such as in our `<Form>` component). Ideally, we'd map our components like:

- `<Form>` contains `<form>` element
- `<Section>` contains `<fieldset>` element
- `<*Question>` contains `<input>` element

client/src/components/IntegerQuestion.jsx
```diff
@@ -0,0 +1,72 @@
...
+      {/* TODO: `Form` should be moved to the section component */}
+      <Form.Field required error={!!validationState}>
+        {/* eslint-disable-next-line jsx-a11y/label-has-associated-control */}
```
marcja, line 53: Why disabled? One way to solve this is by placing the `<input>` inside the `<label>`. See https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/blob/master/docs/rules/label-has-associated-control.md.

client/src/components/IntegerQuestion.jsx
marcja, line 56: Don't use inline styles. Define class, like `class="questionHint"` instead.

client/src/components/IntegerQuestion.jsx
```diff
@@ -0,0 +1,72 @@
...
+  const {questionKey} = props;
+  const a11yId = `integer-question-${questionKey}`;
+
+  const dispatch = useDispatch();
+  const responseValue = useSelector(selectResponses)[questionKey];
+  const questionData = useSelector(selectQuestions)[questionKey];
+  const [validationState, setValidationState] = useState(ValidationState.VALID);
+
+  // TODO: replace with i18next
+  const questionStrings = questionData.resources.en;
```
marcja, line 33: It may make the code easier to read if `s/questionStrings/question/g`.

client/src/components/IntegerQuestion.jsx
```diff
@@ -0,0 +1,72 @@
...
+  return (
+    <Form size='large'>
+      {/* TODO: `Form` should be moved to the section component */}
+      <Form.Field required error={!!validationState}>
+        {/* eslint-disable-next-line jsx-a11y/label-has-associated-control */}
+        <label htmlFor={a11yId}>{questionStrings.text}</label>
+        {!!questionStrings.hint &&
+          <p style={{'font-size': '1rem'}}>{questionStrings.hint}</p>}
+        <Input
+          id={a11yId} type='number' step={1} min={0}
```
marcja, line 58: I think if you contain the input inside the label you can eliminate the id.

client/src/components/IntegerQuestion.jsx
```diff
@@ -0,0 +1,72 @@
...
+          value={responseValue} onChange={handleInput}
+        />
+        {!!validationState &&
+          <Label pointing prompt>{validationState}</Label>}
```
marcja, line 62: Should this be using [setCustomValidity](https://developer.mozilla.org/en-US/docs/Web/API/HTMLObjectElement/setCustomValidity) from the [Constraint validation API](https://developer.mozilla.org/en-US/docs/Web/API/Constraint_validation) instead?

client/src/components/Form.jsx
egracer: #16 specifically stated requiring non-negative integers, but I don't have a problem changing the name if you think it's more clear.

client/src/components/IntegerQuestion.jsx
egracer: Done

client/src/components/IntegerQuestion.jsx
egracer: That seems to be how the tests were written in [this file in #26](https://github.com/ny/maslow/pull/26/files#diff-0fbe6a3900aa269e1deaa5a779cd8dfc). Open to revisiting though that's probably a different PR.

client/src/components/IntegerQuestion.jsx
egracer: Yes, this is temporary just because the rest of the UI components are still being developed, and Semantic UI throws errors if its `Form` component isn't present.

client/src/components/IntegerQuestion.jsx
egracer: Disabled because the linter is looking specifically for an `input` label, but we are using Semantic UI's `Input` component. I will try moving the `Input` component inside the label but I'm skeptical that the linter will be able to accurately identify this case.

client/src/components/IntegerQuestion.jsx
```diff
@@ -0,0 +1,72 @@
...
+
+    // Note: we are always saving the value that the user enters to prevent
+    // inconsistent state between the Redux store and the value displayed in the
+    // input box.
+    dispatch(responsesUpdate({questionKey: questionKey, response: newValue}));
+  };
+
+  return (
+    <Form size='large'>
+      {/* TODO: `Form` should be moved to the section component */}
```
mscharfstein, line 51: will this work? can you have a <Form.Field> not inside a <Form> in a given component?

client/src/components/IntegerQuestion.jsx
mscharfstein: maybe we should have a CSS "hint" class that handles the font size and styling and we can use it for all hints

client/src/components/IntegerQuestion.jsx
egracer: Done, replaced with `/^\d+$/`.

client/src/components/IntegerQuestion.jsx
egracer: Yeah, I'm not a huge fan of this either. It's the only way I've been able to find that allows me to put the hint text in between the label and the input field (see the [Semantic UI page on forms](https://react.semantic-ui.com/collections/form/#shorthand-subcomponent-control)). Unless someone can identify a cleaner pattern, or unless we don't put hint text here, we'll have to keep it.

client/src/components/IntegerQuestion.jsx
egracer: I'm pretty sure it will, yes, as long as Semantic UI's `<Form>` is used in a parent component.

client/src/components/IntegerQuestion.jsx
egracer: Done. I was thinking that `question` is a bit too overloaded of a term but agreed it makes the code more readable.

client/src/components/IntegerQuestion.jsx
egracer: Yeah, unfortunately this is needed. I'll add a comment to clarify.

client/package.json
```diff
@@ -21,7 +21,9 @@
     "react-i18next": "^11.7.0",
     "react-redux": "^7.1.3",
     "react-router-dom": "^5.2.0",
-    "react-scripts": "3.4.1"
+    "react-scripts": "3.4.1",
+    "redux": "^4.0.5",
```
kebrahim, line 25: do we need to explicitly add redux? i thought we already using it without needing this

client/src/components/Form.jsx
egracer: Done. I guess my thinking was -- maybe in the future, a new type of question might allow decimals. But that's an easily solvable future problem.

client/src/components/IntegerQuestion.jsx
egracer: Was not aware that this was a thing, will look into it, thanks!

client/package.json
egracer: There were warnings during compilation about needing to require it as a peer dependency. There are still a few others in there for linting that I wasn't able to resolve.

client/public/index.html
egracer: I did, `semantic-ui-css`, but just discovered I needed to import from the JS, not the HTML. Fixed.

client/src/components/IntegerQuestion.jsx
marcja: Here's one way:

```javascript
dispatch(responsesUpdate(Object.fromEntries([[questionKey, newValue]])));
```

egracer: Nice, thank you. I'm going to leave this as is for expediency per our offline discussion about prioritizing the boolean component, but we can address in a future PR. There's no functional issue, but it would definitely make the code more concise

client/src/components/IntegerQuestion.jsx
egracer: done

egracer: Done

client/src/components/IntegerQuestion.jsx
egracer: Per offline discussion, postponed for a future PR.

client/src/components/NumberQuestion.jsx
```diff
@@ -0,0 +1,81 @@
...
+ *   - questionKey: a unique identifier for the question to be rendered
+ * @return {object} a React component which asks a question and accepts only a
+ * positive integer response.
+ */
+function NumberQuestion(props) {
+  const {questionKey} = props;
+  const a11yId = `integer-question-${questionKey}`;
+
+  const dispatch = useDispatch();
+  const responseValue = useSelector(selectResponses)[questionKey];
```
kebrahim, line 29: how about calling this responseData, to be consistent w/ questionData, so it's clear these are coming from Redux?

client/src/components/NumberQuestion.jsx
```diff
@@ -0,0 +1,81 @@
...
+  INVALID: 'Please enter a whole number greater than or equal to 0.',
+};
+
+/**
+ * @param {object} props
+ *   - questionKey: a unique identifier for the question to be rendered
+ * @return {object} a React component which asks a question and accepts only a
+ * positive integer response.
+ */
+function NumberQuestion(props) {
```
kebrahim, line 24: add TODO to add unit tests

client/src/components/NumberQuestion.jsx
mscharfstein: +1

client/src/index.js
```diff
@@ -1,6 +1,6 @@
 import React from 'react';
 import ReactDOM from 'react-dom';
-import './index.css';
+import '../node_modules/semantic-ui-css/semantic.min.css';
```
mscharfstein, line 3: im confused by this, we shouldnt have to import from node_modules. running yarn add semantic-ui-react and then import the specific elements (import { Button } from 'semantic-ui-react') should be enough. but won't block on this

client/src/components/NumberQuestion.jsx
egracer: Done

client/src/components/NumberQuestion.jsx
egracer: Done

client/src/index.js
egracer: Yeah this confused me too -- see https://react.semantic-ui.com/usage. The `semantic-ui-react` package just adds predefined classes to react-dom comonents. Separately, you have to choose a Semantic UI 'theme' to add styling based on those auto-populated classes.

PR #47, opened 2020-07-17
Add a view reducer slice

mscharfstein:
Works on #9 

View slice handles updating the current section view and well as the validity of each question.

Also remove the counter slice since that is not actually used.


client/src/features/view/viewSlice.js
```diff
@@ -0,0 +1,30 @@
...
+    },
+  },
+});
+
+export const {
+  updateCurrentSectionView,
+  updateResponseValidity,
+} = viewSlice.actions;
+
+export const selectView = (state) => state.view;
```
kebrahim, line 28: should we have individual selectors for the two fields? i'm not sure what's best practice here...

client/src/features/view/viewSlice.js
```diff
@@ -0,0 +1,30 @@
+import {createSlice} from '@reduxjs/toolkit';
+
+export const viewSlice = createSlice({
+  name: 'view',
+  initialState: {
+    currentSection: 1,
```
kebrahim, line 6: should this be set by App.js? i'm honestly not sure of the pros/cons

client/src/features/view/viewSlice.js
```diff
@@ -0,0 +1,30 @@
...
+    responseValidity: {},
+  },
+  reducers: {
+    updateCurrentSectionView: (state, action) => {
+      state.currentSection = action.payload.currentSection;
+    },
+    updateResponseValidity: (state, action) => {
+      for (const [questionKey, isValid] of Object.entries(
```
marcja, line 14: Nit: FWIW, you could use forEach instead:

```javascript
Object.entries(action.payload.responseValidity).forEach(([questionKey, isValid) => {
...
});
```

client/src/features/view/viewSlice.js
```diff
@@ -0,0 +1,30 @@
+import {createSlice} from '@reduxjs/toolkit';
+
+export const viewSlice = createSlice({
+  name: 'view',
+  initialState: {
+    currentSection: 1,
+    responseValidity: {},
+  },
+  reducers: {
+    updateCurrentSectionView: (state, action) => {
```
marcja, line 10: s/updateCurrentSectionView/updateCurrentSection/g?

client/src/features/view/viewSlice.js
mscharfstein: i want to set responseValidity to an empty object so that we don't have to check for its existing in updateResponseValidity. but i can remove currentSection

client/src/features/view/viewSlice.js
mscharfstein: i think its just a matter of preference, have seen it both ways in the tutorial

client/src/features/view/viewSlice.js
mscharfstein: done

client/src/features/view/viewSlice.js
mscharfstein: done

client/src/features/view/viewSlice.js
mscharfstein: set currentSection to 0 based on ping discussion

client/src/features/view/viewSlice.js
```diff
@@ -0,0 +1,30 @@
+import { createSlice } from "@reduxjs/toolkit";
+
+export const viewSlice = createSlice({
+  name: "view",
+  initialState: {
+    currentSection: 0,
```
kebrahim, line 6: add a comment explaining this is the first section (as opposed to no section)

PR #49, opened 2020-07-17
Support i18n-chained-backend

marcja:
Enables translation based on dynamic translation content served from /api/explore/v1/locales.


client/src/components/Form.jsx
```diff
@@ -20,7 +20,7 @@ function Form() {
     }
 
     const renderQuestionKey = (questionKey) => {
-      return <li>{questionKey}</li>;
+      return <li>{t(`catalog:${questionKey}.text`)}</li>;
```
itssumertime, line 23: What is the "catalog" it pulls the "text" field from?

marcja: It is the [i18next namespace](https://www.i18next.com/principles/namespaces).

PR #50, opened 2020-07-17
add translations for sections/questions in 'es'

marcja:


server/src/main/resources/application.yaml
```diff
@@ -36,15 +36,21 @@ maslow:
       SECTION_HOUSEHOLD_INCOME:
         resources:
           en:
-            title: Household Income
+            title: Household income
```
itssumertime, line 39: What's the reason we want to uncapitalize the second word?

marcja: Personal preference. I'm sure it will all get reviewed by the UX writer.

PR #55, opened 2020-07-17
Add i18n for title and button in Summary and Form pages

itssumertime:
Updates i18n for part of Summary page ("back" button, title "Summary") and for the Form page ("Form" title and "submit" button).


client/public/locales/bn/translation.json
```diff
@@ -1,3 +1,5 @@
 {
-  "Form": "ফর্ম"
+  "Back": "প্রত্যাবর্তন",
+  "Form": "ফর্ম",
+  "Summary": "সারসংক্ষেপ"
```
marcja, line 4: Please see [i18next JSON format](https://www.i18next.com/misc/json-format). These should be camel-cased keys, not phrases. Also consider if we should be grouping keys, just as you suggested for catalog strings. For example, "button.back" or "button: { back: "Back" }".

itssumertime: Done. Also added "Submit" button since it seemed to fit in well with the "button" formatting.

PR #57, opened 2020-07-17
add i18n to NumberQuestion

marcja:


client/src/components/NumberQuestion.jsx
```diff
@@ -61,7 +61,8 @@ function NumberQuestion(props) {
         contains an `input` subcomponent, making the lint error inaccurate.
         eslint-disable-next-line jsx-a11y/label-has-associated-control */}
       <label htmlFor={a11yId}>{question.text}</label>
-      {!!question.hint && <p className="questionHint">{question.hint}</p>}
+      {!!question.hint &&
```
egracer, line 64: Will this conditionally hide the hint? I think even if `questionKey` is null, the `p` tag will always be rendered. I ran into this with `BooleanQuestion` and it actually looks ok even if the hint text is blank, so we could just remove the conditional rendering here altogether.

client/src/components/NumberQuestion.jsx
```diff
@@ -2,13 +2,13 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import {useState} from 'react';
 import {useSelector, useDispatch} from 'react-redux';
-import {selectQuestions} from '../features/questions/questionsSlice';
```
mscharfstein, line 5: we might not even need the questions in the redux store at all....

client/src/components/NumberQuestion.jsx
marcja: It's your code, man. I just added the translations. ;)

client/src/components/NumberQuestion.jsx
marcja: Still need to pull the `question.type`.

PR #59, opened 2020-07-17
Added tests for App.js

kebrahim:
Works on #13 #72 

This includes:
- Verifying the API is getting called via axios (FYI, i switched from `fetch` to `axios`, because that seems to be the better supported library for making API calls, especially for older browsers, which we need to support)
- Verifying Redux is being called to store the API data
- Mocking out both axios & Redux & verifying the correct data is being passed back & forth
- Adding VSCode launch configs to debug all React tests & all tests in a single file

This might be overkill for this particular class, but mocking out Redux should be useful for other test classes.


client/src/App.test.js
```diff
@@ -1,16 +1,112 @@
...
+import axios from 'axios';
+import configureMockStore from 'redux-mock-store';
+import thunk from 'redux-thunk';
+
+jest.mock('axios');
+jest.mock('react-i18next', () => ({
+  useTranslation: () => ({t: (key) => key}),
+}));
+
+const mockForm = [
```
marcja, line 14: I think the correct style is to move these manual mocks to a `__mocks__` subdirectory (see https://jestjs.io/docs/en/manual-mocks).

client/src/App.test.js
```diff
@@ -1,16 +1,112 @@
...
+  'questions': mockQuestions,
+  'services': mockServices,
+};
+
+const middlewares = [thunk];
+const mockStore = configureMockStore(middlewares);
 
 // TOOD(kebrahim): Add tests confirming data is stored in Redux
-test('renders form', () => {
+test('calls API', async () => {
```
marcja, line 77: Before we write too many more tests, can we discuss standardizing on the `it(...)` form vs the `test(...)` form? I just find it reads better and encourages better test names.

mscharfstein: i think the "it" uses enzyme (but i could be wrong)

.vscode/launch.json
```diff
@@ -16,6 +16,32 @@
       "sourceMapPathOverrides": {
         "webpack:///src/*": "${webRoot}/*"
       }
-    }
+    },
+    {
+      "name": "Debug All React Tests",
```
mscharfstein, line 21: do these run automatically? can we run these manually?

client/src/App.test.js
egracer: +1 for readability / separation of concerns

client/src/App.test.js
egracer: [`it` is an alias for `test`](https://jestjs.io/docs/en/api.html#testname-fn-timeout). I agree with Mark tests are much easier for me to read with that convention, but if others aren't swayed it's not a dealbreaker for me (pun very much intended 😬)

client/src/Routes.test.js
```diff
@@ -6,6 +6,10 @@ import Routes from './Routes';
 import {Provider} from 'react-redux';
 import store from './app/store';
 
+jest.mock('react-i18next', () => ({
```
egracer, line 9: Does this have to be added to every test file? Or just every test that renders an internationalized string?

kebrahim: the latter

.vscode/launch.json
kebrahim: you can run them directly in VSCode - in the `Run` tab, select the `Debug All React Tests` or `Debug current React file` configurations from the dropdown in the top-left corner

kebrahim:
> this looks great to me! thanks for pushing through this. did you get coverage to work?

no, not yet; will do that separately

client/src/App.test.js
marcja: The linter I just checked in *can* [enforce rules](https://github.com/jest-community/eslint-plugin-jest/blob/master/docs/rules/consistent-test-it.md) about `test` vs `it` (whether inside a `describe` or not), but it is not currently enabled. Personally, I prefer `it` clauses within a `describe` clause, but `test` is fine outside of a `describe` (which is actually [the default for the linter rule](https://github.com/jest-community/eslint-plugin-jest/blob/master/docs/rules/consistent-test-it.md#default-configuration)).

kebrahim: agreed, i think 'describe' & 'it' makes for better readability. updated for these 3 test files. we can update the other test files in a separate PR

client/src/App.test.js
kebrahim: so, i think these are actually test fixtures, not necessarily mocks. it's not like we're mocking out a module here. so, i moved this data to a `__fixtures__` directory & then referenced them here. 

i like this because we can use these fixtures for other tests, as well. thanks for the suggestion!

PR #68, opened 2020-07-20
Connect 'Back' button on SummaryPage

### LINKED ISSUE #10: Create the summary page component

egracer:
Closes #10.

Clicking the back button now properly links to the question form.


client/src/components/SummaryPage.jsx
```diff
@@ -18,9 +19,9 @@ function SummaryPage() {
           {questionKey}: {String(response)}
         </p>
       ))}
-      {/* TODO: currently a noop, needs to be hooked up to the last of the
-      question group pages */}
-      <button>Back</button>
+      <Link to="/">
```
mscharfstein, line 22: eventually we may want to update the "view" state here to indicate which section of the homepage to go back to. lgtm for now

PR #71, opened 2020-07-20
Fix formatter/linter issues raised by marcja/chore/codestyle

marcja:


client/src/components/BooleanQuestion.jsx
```diff
@@ -1,63 +1,73 @@
...
+          name={radioGroupName}
+          checked={responseData === "false"}
+          onChange={handleChange}
         />
-        {showError &&
+        {showError && (
           // TODO: translation
-          <Label pointing prompt>Please select yes or no.</Label>}
+          <Label pointing prompt>
+            {t("Please select yes or no.")}
```
egracer, line 68: Does this need to be an i18n-next key instead of a phrase? There are a few more cases of this in the PR

marcja: Yes, this needs to changed ultimately. Right now, this will actually display "Please select yes or no." in the UI, so it's not worth changing at this point.

PR #76, opened 2020-07-20
Update the format of the responses slice / actions

### LINKED ISSUE #74: Change responsesSlice actions to cleaner format
- Put questionKey directly as the key in the payload (`<questionKey>: <response>`) rather than having a questionKey: `<questionKey>`, response: `<response>`
- Update the use of this action in BooleanQuestion and NumberQuestion

mscharfstein:
Fixes #74 

- Put questionKey directly as the key in the payload (<questionKey>: <response>) rather than having a questionKey: <questionKey>, response: <response>


client/src/components/NumberQuestion.jsx
```diff
@@ -52,7 +52,7 @@ function NumberQuestion(props) {
     // Note: we are always saving the value that the user enters to prevent
     // inconsistent state between the Redux store and the value displayed in the
     // input box.
-    dispatch(updateResponses({ questionKey: questionKey, response: newValue }));
+    dispatch(updateResponses({ questionKey: newValue }));
```
marcja, line 55: You probably meant: 
```js
dispatch(updateResponses({ [questionKey]: newValue })
```

Did you test your change?

mscharfstein: i tested it within the reducer slice test file and it worked fine (but i guess the test is not structured exactly like it is here. i filed another bug to add tests to the FE components). but why do i need the [ ] ?

client/src/features/responses/responsesSlice.js
```diff
@@ -18,14 +18,7 @@ export const responsesSlice = createSlice({
...
     updateResponses: (state, action) => {
-      /**
-       * TODO: update this so that action.payload has the form:
-       *
-       * {'HOUSEHOLD_INCOME': 500}
-       *
-       * vs. currently {questionKey: 'HOUSEHOLD_INCOME', response: 500}
-       */
-      state[action.payload.questionKey] = action.payload.response;
+      Object.assign(state, action.payload.responses);
```
egracer, line 21: @mscharfstein Sumer found a bug, responses aren't being saved in the form on master. I changed this line locally to `action.payload` instead of `action.payload.response` and the form started working. Is that the right way to fix this, or should we change how the payload is passed from the components sending the update?

egracer: This kind of thing will be much easier to catch once I've written tests for the components, so no major concerns for me on the "how do we prevent this moving forward" angle.

PR #84, opened 2020-07-20
add support for extract i18n keys

marcja:
Add a new infrastructure script that will automatically extract all translation keys from codebase and synchronize them across all our translation files. To run the command:

```sh
yarn run i18n:extract
```

Note that locale 'ht' is currently not working due to a [bug in i18next](https://github.com/i18next/i18next/issues/1491). I issued a [PR](https://github.com/i18next/i18next/pull/1492) in i18next to fix it.


client/package.json
```diff
@@ -58,6 +59,9 @@
       "not op_mini all"
     ]
   },
+  "eslintConfig": {
```
kebrahim, line 62: is this needed for i18n extract?

marcja: Oh, good catch. That was a merge error. Fixed.

PR #91, opened 2020-07-21
Adds unit tests for NumberQuestion

### LINKED ISSUE #16: Render integer questions

kebrahim:
Works on #16 

Covered the following test cases:
- Rendering with question text/hint
- Rendering with question text without hint
- Rendering with response pulled from Redux
- Updating input with valid response, including Redux
- Ignoring invalid response (non-integer)
- Validation (I'm assuming Michelle will cover this in her PR)
- a11y (I'll do this in a subsequent PR)


client/src/components/NumberQuestion.test.js
```diff
@@ -0,0 +1,137 @@
...
+    fireEvent.change(numberInput, { target: { value: "hi there" } });
+
+    // Verify the invalid input is not saved in UI or Redux
+    expect(
+      getByTestId("integer-question-HOUSEHOLD_SIZE").children[0].value
+    ).toEqual("");
+    expect(store.getActions()).toHaveLength(0);
+  });
+
+  // TODO(mscharfstein): Add test for validation
```
mscharfstein, line 134: the above test is already a test for validation i think..

client/src/components/NumberQuestion.test.js
```diff
@@ -0,0 +1,137 @@
...
+      if (key != "catalog:ADULTS_65_PLUS.hint") {
+        return key;
+      } else {
+        return null;
+      }
+    },
+  }),
+}));
+
+const middlewares = [thunk];
```
marcja, line 21: What do you use thunk for?

client/src/components/NumberQuestion.test.js
```diff
@@ -0,0 +1,137 @@
...
+    const questionKey = "HOUSEHOLD_SIZE";
+    const store = mockStore({
+      questions: mockQuestions,
+      responses: {},
+    });
+    const { getByText, getByTestId } = render(
+      <Provider store={store}>
+        <NumberQuestion questionKey={questionKey} />
+      </Provider>
+    );
```
marcja, line 36: This is a lot of repetitive overhead for each test. See my PR #94 for an alternative.

client/src/components/NumberQuestion.test.js
```diff
@@ -0,0 +1,137 @@
...
+    expect(getByText("catalog:HOUSEHOLD_SIZE.text")).toBeInTheDocument();
+    expect(getByText("catalog:HOUSEHOLD_SIZE.hint")).toBeInTheDocument();
+    expect(
+      getByTestId("integer-question-HOUSEHOLD_SIZE").children[0].value
```
marcja, line 40: I don't think you need a testid here. Maybe:
```js
getByLabelText("catalog:HOUSEHOLD_SIZE.text").toHaveValue("")
```
[toHaveValue](https://github.com/testing-library/jest-dom#tohavevalue)

client/src/components/NumberQuestion.test.js
```diff
@@ -0,0 +1,137 @@
...
+    ).toEqual("");
+  });
+
+  it("Hides hint label when hint not specified", () => {
+    const questionKey = "ADULTS_65_PLUS";
+    const store = mockStore({
+      questions: mockQuestions,
```
marcja, line 47: Not convinced that using all this real data as mock is necessarily clearer in these simple cases.

client/src/components/NumberQuestion.test.js
```diff
@@ -0,0 +1,137 @@
...
+      responses: {},
+    });
+    const { getByText, queryByText, getByTestId } = render(
+      <Provider store={store}>
+        <NumberQuestion questionKey={questionKey} />
+      </Provider>
+    );
+    expect(getByText("catalog:ADULTS_65_PLUS.text")).toBeInTheDocument();
+    expect(queryByText("catalog:ADULTS_65_PLUS.hint")).toBeNull();
```
marcja, line 56: Better would be:
```js
expect(queryByText("...")).not.toBeInTheDocument()
//or
expect(queryByText("...")).not.toBeVisible()
```

client/src/components/NumberQuestion.test.js
```diff
@@ -0,0 +1,137 @@
...
+    const { getByTestId } = render(
+      <Provider store={store}>
+        <NumberQuestion questionKey={questionKey} />
+      </Provider>
+    );
+
+    // Update number input to a valid integer
+    const numberInput = getByTestId("integer-question-HOUSEHOLD_SIZE")
+      .children[0];
+    expect(numberInput.value).toEqual("");
```
marcja, line 95: See above. Better to use the .toHaveValue matcher.

client/src/components/NumberQuestion.test.js
```diff
@@ -0,0 +1,137 @@
...
+    fireEvent.change(numberInput, { target: { value: "3" } });
+    expect(
+      getByTestId("integer-question-HOUSEHOLD_SIZE").children[0].value
+    ).toEqual("3");
```
marcja, line 99: What is the point of this assertion? On the line above, you're changing the state of the HTML element and immediately validating that the HTML element still has that value. How could this fail?

client/src/components/NumberQuestion.test.js
kebrahim: good point, removed TODO

client/src/components/NumberQuestion.test.js
kebrahim: removed

client/src/components/NumberQuestion.test.js
kebrahim: thanks for the tips. i cleaned it up tremendously, using both yours & ethan's PRs as guides. it looks much better now :)

client/src/components/NumberQuestion.test.js
kebrahim: removed

kebrahim:
> This all looks good, could you also add tests checking that negative & decimal values show the error state?

done

client/src/components/NumberQuestion.test.js
kebrahim: done

client/src/components/NumberQuestion.test.js
kebrahim: done

client/src/components/NumberQuestion.test.js
kebrahim: done

client/src/components/NumberQuestion.test.js
kebrahim: removed

PR #92, opened 2020-07-21
Move QuestionForm section logic into Section component

mscharfstein:
Works on #14 

Moves the logic for rendering sections from QuestionForm to Section. Hardcodes a given "current page" value, @marcja 's navigation PR will handle this part.

Missing tests right now; waiting for @marcja 's navigation PR with tests for reference. Also missing Semantic theming to look like NYS.

Form looks the same, just rendering a single section at a time:
image


client/src/components/Section.jsx
```diff
@@ -12,80 +15,37 @@ import PropTypes from "prop-types";
...
-    <Form.Group>
-      <legend data-testid="section-header">
-        Step {props.currentStep} of {props.totalSteps}
-      </legend>
-      {validationErrors}
-      {questionElements}
-      <button>Continue</button>
-      <button>Go back to previous page</button>
-    </Form.Group>
+    <Container>
```
marcja, line 34: A section should not have a container. That is provided by the Form. You can use a React.Fragment instead, a la:

```js
return (
  <>
    {Object.keys(
     // ...
   )]
  </>
);
```

mscharfstein: done

PR #93, opened 2020-07-21
Added axe-core test for App.js

kebrahim:
Works on #8 

Adds the `axe-core` library to our tests which checks for a11y violations, specifically all of those mentioned here: https://github.com/dequelabs/axe-core/blob/develop/doc/rule-descriptions.md 

In order to use it, we need to use the library in each of our component tests.

Note that I discovered that `jest-axe` is a testing library that wraps around `axe-core`, allowing for testing for a11y without requiring any changes to the application. You can compare this to using `axe-core` directly, which I implemented in maslow-experimental: https://github.com/ny/maslow-experimental/pull/20

I like this better. :)


marcja:
Can you provide some (links to?) documentation as to what this actually does? What's constitutes a violation?

kebrahim:
> Can you provide some (links to?) documentation as to what this actually does? What's constitutes a violation?

done, updated PR description

PR #95, opened 2020-07-21
Validate that none of the questions are left empty

mscharfstein:
- Disables submit button until all questions (boolean and otherwise) are filled out
- Shows error state for the boolean questions that are not filled out when you try to submit
- Removes response validity section of the view slice (not actually needed; can check that answer is filled out in the responses slice, and the number validation happens directly when entering numbers into the form and only needs to be accessed by the NumberQuestion form)
- TODO: Add tests

When you try to submit without filling out boolean or number questions:
image


client/src/components/BooleanQuestion.jsx
```diff
@@ -37,7 +41,7 @@ function BooleanQuestion(props) {
   };
 
   return (
-    <Form.Field required error={showError}>
+    <Form.Field required error={showError && !validity}>
```
marcja, line 44: This seems like it would be awkward to test. If I wanted to test the component in error state, I can't just pass in a prop anymore. It seems like this component is unsure whether it is purely a presentation component or a logic component.

client/src/components/QuestionForm.jsx
```diff
@@ -20,6 +21,9 @@ function QuestionForm() {
   const { t } = useTranslation();
   const history = useHistory();
 
+  const [showError, setShowError] = useState(false);
+  const responseData = useSelector(selectResponses);
```
marcja, line 25: s/responseData/responses/g

client/src/components/QuestionForm.jsx
```diff
@@ -51,7 +61,19 @@ function QuestionForm() {
...
+    let canSubmit = true;
+    const responseKeys = Object.keys(responseData);
+    const questionKeys = Object.keys(questions);
+    for (let i = 0; i < questionKeys.length; i++) {
+      if (responseKeys[questionKeys[i]] == null) {
+        canSubmit = false;
+        setShowError(true);
+        break;
+      }
+    }
```
marcja, line 73: Aren't we just says that every questionKey can also be found as a responseKey? So just:

```js
const valid = questionsKeys.every((key) => responseKeys.includes(key));
```

A negligible performance improvement would be:

```js
const responseKeys = new Set(Object.keys(responses));
...
const valid = questionsKeys.every((key) => responseKeys.has(key));
```

client/src/components/QuestionForm.jsx
```diff
@@ -51,7 +61,19 @@ function QuestionForm() {
...
+    if (canSubmit) {
+      history.push("/summary");
```
marcja, line 75: Another way to do this is to have the button be a proper link:

```js
<Link to="/summary">
  <Button type="submit" ... />
</Link
```

And simply return a boolean from `handleSubmit`:

```js
return canSubmit;
```

If you return false from an onSubmit handler, it doesn't navigate.

client/src/components/QuestionForm.jsx
mscharfstein: done

client/src/components/QuestionForm.jsx
mscharfstein: done

client/src/components/QuestionForm.test.js
```diff
@@ -3,6 +3,7 @@ import { render } from "@testing-library/react";
 import { Provider } from "react-redux";
 import store from "../app/store";
 import Form from "./QuestionForm.jsx";
+import { BrowserRouter as Router } from "react-router-dom";
```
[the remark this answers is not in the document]
mscharfstein: this is outdated, not using anymore. same with the others

client/src/App.test.js
```diff
@@ -35,9 +36,11 @@ describe("App", () => {
     });
 
     const { getByText } = render(
-      <Provider store={store}>
-        <App />
-      </Provider>
+      <Router>
```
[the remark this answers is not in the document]
mscharfstein: outdated

client/src/App.test.js
```diff
@@ -5,6 +5,7 @@ import App from "./App";
 import axios from "axios";
 import configureMockStore from "redux-mock-store";
 import thunk from "redux-thunk";
+import { BrowserRouter as Router } from "react-router-dom";
```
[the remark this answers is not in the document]
mscharfstein: outdated

client/src/components/BooleanQuestion.jsx
```diff
@@ -24,6 +24,10 @@ function BooleanQuestion(props) {
   const dispatch = useDispatch();
   const { t } = useTranslation();
   const responseData = useSelector(selectResponses)[questionKey];
+  let validity = false;
```
kebrahim, line 27: i'm confused about what this variable represents. maybe it's the variable name? e.g. if showError is true but there's a response in Redux, then we wouldn't show an error. is that expected behavior?

client/src/components/NumberQuestion.jsx
```diff
@@ -1,20 +1,14 @@
...
 import "../css/questions.css";
 import { useTranslation } from "react-i18next";
 
-// TODO: translations, move into Redux after agreement on validation strategy
-const ValidationState = {
-  VALID: "",
-  INVALID: "Please enter a whole number greater than or equal to 0.",
-};
+// TODO: translation
+const invalid = "Please enter a whole number greater than or equal to 0.";
```
kebrahim, line 11: s/invalid/invalidInputMessage

client/src/components/NumberQuestion.jsx
```diff
@@ -42,11 +35,12 @@ function NumberQuestion(props) {
 
   const handleInput = (event) => {
     const newValue = event.target.value;
+    setNumber(newValue);
     const NUMBER_PATTERN = /^\d+$/;
     if (NUMBER_PATTERN.test(newValue)) {
```
kebrahim, line 40: this if...else clause could be replaced with:
```
setValidity(NUMBER_PATTERN.test(newValue));
```

client/src/components/NumberQuestion.jsx
```diff
@@ -26,14 +20,13 @@ function NumberQuestion(props) {
...
+  const { questionKey, showErrorForNoEntry } = props;
   const a11yId = `integer-question-${questionKey}`;
 
   const dispatch = useDispatch();
-  const responseData = useSelector(selectResponses)[questionKey];
 
-  // TODO: replace with an HTML and/or Redux validation library
-  const [validationState, setValidationState] = useState(ValidationState.VALID);
+  const [number, setNumber] = useState(undefined);
+  const [isValid, setValidity] = useState(true);
```
kebrahim, line 29: how about isInputValid/setInputValid?

client/src/components/NumberQuestion.jsx
```diff
@@ -1,20 +1,14 @@
...
-// TODO: translations, move into Redux after agreement on validation strategy
-const ValidationState = {
-  VALID: "",
-  INVALID: "Please enter a whole number greater than or equal to 0.",
-};
+// TODO: translation
+const invalid = "Please enter a whole number greater than or equal to 0.";
 
 /**
  * @param {object} props
```
kebrahim, line 14: update comment explaining what props are passed in

client/src/components/NumberQuestion.jsx
```diff
@@ -26,14 +20,13 @@ function NumberQuestion(props) {
   // TODO: add unit tests
   const { t } = useTranslation();
 
-  const { questionKey } = props;
+  const { questionKey, showErrorForNoEntry } = props;
   const a11yId = `integer-question-${questionKey}`;
 
   const dispatch = useDispatch();
-  const responseData = useSelector(selectResponses)[questionKey];
```
kebrahim, line 33: why did you remove the call to selectResponses? don't we need the response information to pre-populate the input?

client/src/components/QuestionForm.jsx
```diff
@@ -29,23 +32,35 @@ function QuestionForm() {
...
-      <Section sectionKey={sectionKey} questionKeys={sectionInfo[sectionKey]} />
+      <Section
+        sectionKey={sectionKey}
+        questionKeys={sectionInfo[sectionKey]}
+        showError={showError}
+      />
     );
   };
 
   const handleSubmit = () => {
```
kebrahim, line 43: mind adding a comment explaining what's happening in this method?

client/src/components/Section.jsx
```diff
@@ -15,17 +15,29 @@ import PropTypes from "prop-types";
...
   const questions = useSelector(selectQuestions);
   const { t } = useTranslation();
 
   const renderQuestion = (questionKey) => {
     const questionMetadata = questions[questionKey];
     if (questionMetadata.type === "NUMBER") {
-      return <NumberQuestion key={questionKey} questionKey={questionKey} />;
+      return (
+        <NumberQuestion
+          key={questionKey}
```
kebrahim, line 28: where is the `key` prop used in NumberQuestion? it seems like it's superfluous & also we're passing the same value to two different keys.

client/src/components/Section.jsx
```diff
@@ -15,17 +15,29 @@ import PropTypes from "prop-types";
...
+          questionKey={questionKey}
+          showErrorForNoEntry={showError}
```
kebrahim, line 30: is it true that `showErrorForNoEntry` will be true even if all numeric inputs are correct & the user just didn't input a boolean field? is that intended?

client/src/components/Section.jsx
```diff
@@ -15,17 +15,29 @@ import PropTypes from "prop-types";
...
+        />
+      );
     } else if (questionMetadata.type === "BOOLEAN") {
-      return <BooleanQuestion key={questionKey} questionKey={questionKey} />;
+      return (
+        <BooleanQuestion
+          key={questionKey}
```
kebrahim, line 36: similar question about this `key` field not being used

client/src/components/BooleanQuestion.jsx
mscharfstein: that's expected. showError indicates that if no response is selected, we should show the error. i updated that variable name to make it clearer.

client/src/components/Section.jsx
mscharfstein: this was a javascript lint error - you are supposed to provide a key prop when you have multiple children

client/src/components/Section.jsx
mscharfstein: same

client/src/components/Section.jsx
mscharfstein: yes, but the error won't be shown for the number field if there is a number defined (line 55 of NumberQuestion)

client/src/components/NumberQuestion.jsx
mscharfstein: thanks you're totally right. i was trying to replace with local state but i forgot to initialize the local state with the response value

client/src/components/NumberQuestion.jsx
```diff
@@ -56,7 +55,10 @@ function NumberQuestion(props) {
   };
 
   return (
-    <Form.Field required error={!!validationState}>
+    <Form.Field
+      required
+      error={!isInputValid || (showErrorForNoEntry && number === undefined)}
```
marcja, line 60: nit: refactor to extract expression to function

client/src/components/NumberQuestion.jsx
```diff
@@ -67,12 +69,12 @@ function NumberQuestion(props) {
         type="number"
         step={1}
         min={0}
-        value={responseData}
+        value={number}
         onChange={handleInput}
       />
-      {!!validationState && (
+      {(!isInputValid || (showErrorForNoEntry && number === undefined)) && (
```
marcja, line 75: use refactored function from above

client/src/components/NumberQuestion.jsx
```diff
@@ -1,39 +1,41 @@
...
 import PropTypes from "prop-types";
 
-import { useSelector, useDispatch } from "react-redux";
+import { useDispatch, useSelector } from "react-redux";
 import {
-  updateResponses,
   selectResponses,
+  updateResponses,
 } from "../features/responses/responsesSlice";
+
```
kebrahim, line 9: remove?

client/src/components/NumberQuestion.jsx
```diff
@@ -1,39 +1,41 @@
...
-  const { questionKey } = props;
+  const { questionKey, showErrorForNoEntry } = props;
   const a11yId = `integer-question-${questionKey}`;
 
   const dispatch = useDispatch();
   const responseData = useSelector(selectResponses)[questionKey];
 
-  // TODO: replace with an HTML and/or Redux validation library
-  const [validationState, setValidationState] = useState(ValidationState.VALID);
+  const [number, setNumber] = useState(responseData);
```
kebrahim, line 37: `number` is kind of vague, now that i think about it more. how about `answer` or `numberAnswer`?

client/src/components/NumberQuestion.jsx
kebrahim, line 60: could we use the `answered` boolean like you did in the BooleanQuestion.jsx? that was really easy to read/understand

client/src/components/Section.jsx
```diff
@@ -15,17 +15,29 @@ import PropTypes from "prop-types";
  * @return {string} The rendered component.
  */
 function Section(props) {
-  const { sectionKey, questionKeys } = props;
+  const { sectionKey, questionKeys, showError } = props;
```
kebrahim, line 18: mind adding javadoc for each of the props here, like you did in the other components? i find that really useful

client/src/components/NumberQuestion.jsx
marcja: Or, perhaps "response"?

Issue #15, opened 2020-07-14
Render boolean questions


tariqajyusuf:
Boolean (Yes/No) questions need to have:

- [x] The question title
- [x] Any hint text provided
- [x] Radio buttons for yes and no
- [x] Translations for yes/no
- [x] Unit tests
- [ ] Accessibility

kebrahim:
Tests should include (from https://github.com/ny/maslow/issues/72):
- check hint displayed and not displayed (if exists or not)
- check that the right actions are dispatched to the state

egracer:
I think it's also worth a very small integration test to confirm that the value was saved to the store as expected. And that the component loads data from the store as expected. Otherwise it's possible to modify the reducer contract and all of the component tests would still pass.

PR #94, opened 2020-07-21
Update Form component and incorporate navigation

marcja:
This is a draft. It's not done.

PR #96, opened 2020-07-21
i18n for question errors & testing BooleanQuestion

egracer:
Works on #15

Big item: tests for Boolean Question

Small items:

1. i18n for error messages in both the BooleanQuestion & NumberQuestion component.
2. The server now elides `null` fields when serializing JSON. This prevents questions without hints from having `hint: null` stored in Redux, allowing `react-i18next` to identify when a translation is missing for a given string and conditionally render hint text accordingly.


server/src/main/java/gov/ny/its/hs/maslow/model/QuestionResources.java
```diff
@@ -1,41 +1,36 @@
...
 public class QuestionResources {
-  @NotBlank
-  private String text;
+  @NotBlank private String text;
 
   private String hint;
 
   /**
    * Gets the display text of the question.
+   *
```
kebrahim, line 16: can you remove this blank line & the unused @return?

client/src/components/BooleanQuestion.test.js
```diff
@@ -0,0 +1,100 @@
+import React from "react";
+import { render, fireEvent } from "@testing-library/react";
+import { Provider } from "react-redux";
+import configureMockStore from "redux-mock-store";
+import { mockQuestions } from "../__fixtures__/catalog";
+import BooleanQuestion from "./BooleanQuestion.jsx";
+
+// Mock out i18next, ensuring one boolean question does not have hint text
+jest.mock("react-i18next", () => ({
+  useTranslation: () => ({
```
kebrahim, line 10: this can be a TODO, but i think we should switch this i18n mocking to use the renderWith.js file that marc created in his PR: https://github.com/ny/maslow/pull/94/files#diff-f29bf1ac5dfb9479d059403b8866b19eR11-R32

client/src/components/BooleanQuestion.test.js
```diff
@@ -0,0 +1,100 @@
...
+    let mockUi = null;
+
+    beforeEach(() => {
+      mockUi = setup("IS_EMPLOYED");
+    });
+
+    it("initially renders both 'Yes' and 'No' options", () => {
+      const { getByText, queryByText } = mockUi;
+
+      expect(getByText("catalog:IS_EMPLOYED.text")).toBeInTheDocument();
```
kebrahim, line 54: based on https://testing-library.com/docs/guide-which-query  is it possible to switch these to getByRole or getByLabelText instead?

client/src/components/BooleanQuestion.test.js
```diff
@@ -0,0 +1,100 @@
...
+      exists: (key) => {
+        return key != "catalog:IS_EMPLOYED.hint";
+      },
+    },
+  }),
+}));
+
+const mockStore = configureMockStore();
+
+function setup(questionKey, mockResponses, showError) {
```
kebrahim, line 28: i'd also recommend creating another method called `setupWithError(questionKey, mockResponses)` that calls `setup(questionKey, mockResponses, true)`. otherwise, when you call this below (e.g. line 76), it's not clear what the `true` corresponds to. however, if you called `setupWithError(questionKey, {})`, then it's a lot more obvious.

client/src/components/BooleanQuestion.test.js
```diff
@@ -0,0 +1,100 @@
...
+    it("renders the response from Redux", () => {
+      const questionKey = "IS_EMPLOYED";
+      const { getByLabelText } = setup(questionKey, { [questionKey]: "true" });
+
+      expect(getByLabelText("radio.yes")).toBeChecked();
+      expect(getByLabelText("radio.no")).not.toBeChecked();
+    });
+  });
+
+  it("renders hint text if it exists", () => {
```
kebrahim, line 92: mind also adding in an a11y test? see https://github.com/ny/maslow/blob/master/client/src/App.test.js#L65-L83 for an example

client/src/components/BooleanQuestion.test.js
```diff
@@ -0,0 +1,100 @@
...
+        return key != "catalog:IS_EMPLOYED.hint";
+      },
+    },
+  }),
+}));
+
+const mockStore = configureMockStore();
+
+function setup(questionKey, mockResponses, showError) {
+  const store = mockStore({
```
kebrahim, line 29: another TODO, but we should probably use the same Redux store configuration that marc uses in his PR: https://github.com/ny/maslow/pull/94/files#diff-f29bf1ac5dfb9479d059403b8866b19eR34

client/src/components/BooleanQuestion.test.js
```diff
@@ -0,0 +1,100 @@
...
+import { Provider } from "react-redux";
+import configureMockStore from "redux-mock-store";
+import { mockQuestions } from "../__fixtures__/catalog";
+import BooleanQuestion from "./BooleanQuestion.jsx";
+
+// Mock out i18next, ensuring one boolean question does not have hint text
+jest.mock("react-i18next", () => ({
+  useTranslation: () => ({
+    t: (key) => {
+      if (key != "catalog:IS_EMPLOYED.hint") {
```
marcja, line 12: s/!=/!==/g

server/src/main/java/gov/ny/its/hs/maslow/model/QuestionResources.java
```diff
@@ -1,41 +1,36 @@
...
 
+import com.fasterxml.jackson.annotation.JsonInclude;
+import com.fasterxml.jackson.annotation.JsonInclude.Include;
 import javax.validation.constraints.*;
 
-/**
- * Defines the resource bundle for text that needs to be translated for a question.
- */
+/** Defines the resource bundle for text that needs to be translated for a question. */
+@JsonInclude(Include.NON_NULL) // Ensures only non-null fields are serialized to JSON.
```
marcja, line 8: Nice!

client/src/components/BooleanQuestion.test.js
kebrahim: sounds good, please just leave a TODO

egracer: Done

PR #99, opened 2020-07-22
add testing helper: renderWith

marcja:
Provides a testing helper called `renderWith` that wraps `render` from the React Testing Library with configuration specific to our project. Should be used everywhere instead of `render`.


client/src/App.test.js
```diff
@@ -1,16 +1,13 @@
-import React from "react";
-import { render, waitFor } from "@testing-library/react";
-import { Provider } from "react-redux";
-import App from "./App";
+import { screen, waitFor } from "@testing-library/react";
 import axios from "axios";
+import { axe, toHaveNoViolations } from "jest-axe";
+import React from "react";
 import configureMockStore from "redux-mock-store";
 import thunk from "redux-thunk";
```
kebrahim, line 6: i realized that you can remove this & just use the following on line 15:
```
```

client/src/Routes.test.js
```diff
@@ -1,35 +1,34 @@
...
-          <Routes />
-        </MemoryRouter>
-      </Provider>
-    );
-    expect(getByText(/Summary/i)).toBeInTheDocument();
+    const options = {
+      routerOptions: {
+        initialEntries: ["/summary"],
+      },
+      state: {
```
kebrahim, line 20: do we need all these Redux state fields to be initialized? SummaryPage only uses responses

client/src/util/testing.js
```diff
@@ -0,0 +1,91 @@
...
+      useSuspense: false,
+    },
+    resources: { en: { translations: {} } },
+  });
+
+// Configure redux with a mock. 'redux-thunk' is middleware that it is included
+// by default with redux-toolkit, so I'm including it here to ensure that our
+// testing configuration doesn't differ to greatly from our standard.
+
+const mockStore = configureStore([thunk]);
```
kebrahim, line 31: FWIW, ethan & i haven't been using `thunk` in our tests & there hasn't been an issue. i'm honestly not sure why it's necessary, but consider removing it if the tests pass w/out it

client/src/util/testing.js
```diff
@@ -0,0 +1,91 @@
...
+    interpolation: false,
+    react: {
+      useSuspense: false,
+    },
+    resources: { en: { translations: {} } },
+  });
+
+// Configure redux with a mock. 'redux-thunk' is middleware that it is included
+// by default with redux-toolkit, so I'm including it here to ensure that our
+// testing configuration doesn't differ to greatly from our standard.
```
kebrahim, line 29: s/to/too

client/src/util/testing.js
marcja: As I said, it is the default middleware of redux-toolkit. I kept it here to match. Don't remove.

client/src/App.test.js
marcja: No, it should be kept. See comment in testing.js.

client/src/util/testing.js
marcja: fixed

client/src/Routes.test.js
marcja: Good call. Fixed.

PR #100, opened 2020-07-22
Adds currency formatting for NumberQuestion

### LINKED ISSUE #70: Handle currency inputs
Handle Currency in NumberQuestion:

- Pass in prop to indicate if NumberQuestion should display currency
- If so,
      - Display currency ($)
      - [optional] Make the increment for the input (when you press the up/down arrows) be more than 1 for CurrencyQuestions (maybe 50 or 100?)

tariqajyusuf:
Fixes #70


client/src/components/NumberQuestion.jsx
```diff
@@ -16,9 +16,16 @@ const ValidationState = {
   INVALID: "Please enter a whole number greater than or equal to 0.",
 };
 
+NumberQuestion.AnswerTypes = {
```
marcja, line 19: You should consider using the same enum values that the server returns. So:

```js
const NumberQuestion.AnswerTypes = {
  NUMBER: "NUMBER",
  CURRENCY: "CURRENCY"
}
```

client/src/components/NumberQuestion.jsx
```diff
@@ -16,9 +16,16 @@ const ValidationState = {
...
+  DEFAULT: "default",
+  CURRENCY: "currency",
+};
+
 /**
  * @param {object} props
  *   - questionKey: a unique identifier for the question to be rendered
+ *   - answerType: Determines how we format the number either as a "default" or
```
marcja, line 27: Should probably be questionType (at least that's how it's named in the API).

client/src/components/NumberQuestion.jsx
```diff
@@ -39,6 +46,14 @@ function NumberQuestion(props) {
     text: t(`catalog:${questionKey}.text`),
     hint: t(`catalog:${questionKey}.hint`),
   };
+  const formatIcon =
+    props.answerType === NumberQuestion.AnswerTypes.CURRENCY
+      ? "dollar sign"
+      : "";
+  const formatIconPosition =
+    props.answerType !== NumberQuestion.AnswerTypes.DEFAULT ? "left" : null;
```
marcja, line 54: Reverse the clauses to match above? So:

```js
const formatIconPosition =
  props.answerType === NumberQuestion.AnswerTypes.CURRENCY ? "left" : null;
```

client/src/components/NumberQuestion.jsx
```diff
@@ -64,8 +79,10 @@ function NumberQuestion(props) {
       {!!question.hint && <p className="questionHint">{question.hint}</p>}
       <Input
         id={a11yId}
+        icon={formatIcon}
+        iconPosition={formatIconPosition}
         type="number"
-        step={1}
+        step={formatStep}
```
marcja, line 85: FYI, this will make it impossible for someone to type "1234". Is that desired?

client/src/components/NumberQuestion.test.js
```diff
@@ -0,0 +1,46 @@
+import React from "react";
+import { render } from "@testing-library/react";
+import { Provider } from "react-redux";
+import store from "../app/store";
+import NumberQuestion from "./NumberQuestion.jsx";
+
+jest.mock("react-i18next", () => ({
```
marcja, line 7: If you use "renderWith", you will no longer need this mock.

client/src/components/NumberQuestion.test.js
```diff
@@ -0,0 +1,46 @@
...
+  useTranslation: () => ({ t: (key) => key }),
+}));
+
+// TODO: Add tests confirming data from Redux is displayed
+describe("NumberQuestion", () => {
+  it("renders", () => {
+    const { container } = render(
```
marcja, line 14: Using `renderWith`, it's just:

```js
renderWith(<NumberQuestion />);
```

client/src/components/NumberQuestion.test.js
```diff
@@ -0,0 +1,46 @@
...
+      <Provider store={store}>
+        <NumberQuestion />
+      </Provider>
+    );
+
+    expect(container.getElementsByTagName("input")).toHaveLength(1);
```
marcja, line 20: You should not use queries on container directly like this unless there is no other option. See [guidance](https://testing-library.com/docs/guide-which-query). You should prefer the facilities in React Testing Library, such as:

```js
expect(screen.getByLabelText(...)).toBeInTheDocument();
```

client/src/components/NumberQuestion.test.js
```diff
@@ -0,0 +1,46 @@
...
+    expect(container.getElementsByClassName("dollar sign")).toHaveLength(0);
+  });
+  it("renders with explicit default answer type", () => {
```
marcja, line 23: Add newline above.

client/src/components/NumberQuestion.test.js
```diff
@@ -0,0 +1,46 @@
...
+import store from "../app/store";
+import NumberQuestion from "./NumberQuestion.jsx";
+
+jest.mock("react-i18next", () => ({
+  useTranslation: () => ({ t: (key) => key }),
+}));
+
+// TODO: Add tests confirming data from Redux is displayed
+describe("NumberQuestion", () => {
+  it("renders", () => {
```
marcja, line 13: You will save yourself a bunch of headache by habitually adding the `async` keyboard before the lambda in `it` or `test` functions. For example:

```js
it("renders", async () => {...});
```

client/src/components/NumberQuestion.jsx
tariqajyusuf: Makes sense, done.

client/src/components/NumberQuestion.jsx
tariqajyusuf: Done

client/src/components/NumberQuestion.jsx
tariqajyusuf: I don't think that would be correct, because default means we don't want to show anything and so we should not use this attribute. This mainly defensive against the possibility we add other kinds of question types.

client/src/components/NumberQuestion.jsx
tariqajyusuf: I believe this is intended as part of the spec in [#70](https://github.com/ny/maslow/issues/70). It would be significantly more complicated not to use step and I don't think there's a compelling case to support more fine granularity.

mscharfstein: i added that in #70 but didn't realize that you wouldn't be allowed to type whatever you wanted (was just thinking if you wanted to use the arrows to change it, itd be bigger increments) - i guess we could loop in Malika as to whether we should support whatever number or just increments of 100. i don't feel strongly

kebrahim: +1 to asking Malika

marcja: Personally, I think this is an example of a UI trying to be too smart. It's the analogous to having a validation error pop up for the user, "You typed 1234 but the number must end in 00". Like, really?

tariqajyusuf: I'll just undo this for now.

client/src/components/NumberQuestion.test.js
tariqajyusuf: Agreed, I need to use DOM based ones due to the class name ones.

client/src/components/NumberQuestion.jsx
```diff
@@ -39,6 +46,12 @@ function NumberQuestion(props) {
     text: t(`catalog:${questionKey}.text`),
     hint: t(`catalog:${questionKey}.hint`),
   };
+  const formatIcon =
+    props.questionType === NumberQuestion.QuestionTypes.CURRENCY
+      ? "dollar sign"
+      : "";
+  const formatIconPosition =
+    props.questionType !== NumberQuestion.QuestionTypes.DEFAULT ? "left" : null;
```
kebrahim, line 54: any particular reason why you checked for CURRENCY above & then for !DEFAULT here? i'd say make them consistent, if possible

client/src/components/NumberQuestion.test.js
```diff
@@ -1,25 +1,14 @@
...
-      } else {
-        return null;
-      }
-    },
-  }),
-}));
-
 const mockStore = configureMockStore();
 
 // TODO: Use renderWith.js
```
kebrahim, line 14: remove TODO

client/src/components/NumberQuestion.test.js
```diff
@@ -1,25 +1,14 @@
...
-      if (key != "catalog:HOUSEHOLD_INCOME.hint") {
-        return key;
-      } else {
-        return null;
-      }
-    },
-  }),
-}));
-
 const mockStore = configureMockStore();
```
kebrahim, line 12: renderWith does this, so you don't need to keep it here - https://github.com/ny/maslow/blob/master/client/src/util/testing.js#L31

client/src/components/NumberQuestion.test.js
```diff
@@ -28,27 +17,49 @@ function setupUI(questionKey, mockResponses) {
...
   return { store: store, ...mockUi };
 }
 
 describe("NumberQuestion", () => {
   let mockUi = null;
 
+  describe("conditional formatting", () => {
+    it("renders with explicit default answer type", async () => {
+      const { container } = renderWith(
+        <Provider store={store}>
```
kebrahim, line 32: you don't need to pass the Provider to renderWith. it handles that

client/src/components/NumberQuestion.test.js
```diff
@@ -164,25 +174,4 @@ describe("NumberQuestion", () => {
       });
     });
   });
-
-  describe("without hint text", () => {
```
kebrahim, line 168: mind adding a TODO to add back in this test?

client/src/components/NumberQuestion.test.js
```diff
@@ -141,7 +151,7 @@ describe("NumberQuestion", () => {
 
     it("renders number response from Redux", () => {
       const { getByLabelText } = mockUi;
-      expect(getByLabelText("catalog:HOUSEHOLD_SIZE.text")).toHaveValue(7);
+      expect(getByLabelText("HOUSEHOLD_SIZE.text")).toHaveValue(7);
     });
 
     it("has no a11y violations", async () => {
```
kebrahim, line 157: mind either adding a keyboard-tab-navigation test or leaving a TODO? e.g. https://github.com/ny/maslow/blob/master/client/src/components/Form.test.js#L308

client/src/components/NumberQuestion.jsx
mscharfstein: +1

client/src/components/NumberQuestion.test.js
```diff
@@ -28,27 +17,49 @@ function setupUI(questionKey, mockResponses) {
...
+          <NumberQuestion questionType={NumberQuestion.QuestionTypes.DEFAULT} />
+        </Provider>
+      );
+
+      expect(container.getElementsByClassName("dollar sign")).toHaveLength(0);
+    });
+
+    it("renders with currency answer type", async () => {
+      const { container } = renderWith(
+        <Provider store={store}>
```
mscharfstein, line 42: ditto with above, dont need provider in render with

client/src/components/NumberQuestion.jsx
tariqajyusuf: I mentioned this in Marc's comment, the reason for this is that this applies if any icons are shown. If we add another icon type in the future we'd have to make it DEFAULT. So it's more expressive of intent.

client/src/components/NumberQuestion.test.js
tariqajyusuf: Done.

kebrahim:
> you also need to update Section to pass in the answerType prop

good point about updating Section. please also update the correspond test in Section.test.js

client/src/components/NumberQuestion.test.js
tariqajyusuf: Done.

client/src/components/NumberQuestion.test.js
tariqajyusuf: My latest change has this still in and working.

tariqajyusuf:
> > you also need to update Section to pass in the answerType prop
> 
> good point about updating Section. please also update the correspond test in Section.test.js

Section.test.js is not in the repository right now. Let's punt this to another PR.

client/src/components/NumberQuestion.jsx
mscharfstein: i think if we add a new icon, part of adding it would be adding support here. so either setting it to default or to a different type. i personally think it would be better to check consistently and if we add something in the future, make the necessary updates

client/src/components/NumberQuestion.test.js
```diff
@@ -1,59 +1,74 @@
...
+
   describe("without a response", () => {
     beforeEach(() => {
       mockUi = setupUI("HOUSEHOLD_SIZE");
     });
 
     it("renders empty number input", () => {
       const { getByText, getByLabelText } = mockUi;
-      expect(getByText("catalog:HOUSEHOLD_SIZE.text")).toBeInTheDocument();
-      expect(getByText("catalog:HOUSEHOLD_SIZE.hint")).toBeInTheDocument();
```
kebrahim, line 55: mind adding the check for hint back in?

client/src/components/Section.jsx
```diff
@@ -38,7 +38,7 @@ function Section(props) {
           key={questionKey}
           questionKey={questionKey}
           showErrorForNoEntry={showErrorForNoEntry}
-          // TODO(tariq): Pass the prop indicating the Number type here
+          questionType={questionMetadata.type}
```
kebrahim, line 41: add TODO to add test

mscharfstein, line 41: i'm not sure this will work (did it in the UI?) because the type passed here will just be "CURRENCY" or "INTEGER", whereas the types you handle in Section are NumberQuestion.QuestionTypes.CURRENCY or NumberQuestion.QuestionTypes.DEFAULT. might need to either update what's passed or how the questionType is handled

client/src/components/NumberQuestion.jsx
```diff
@@ -58,6 +74,8 @@ function NumberQuestion(props) {
       )}
       <Input
         id={a11yId}
+        icon={formatIcon}
+        iconPosition={formatIconPosition}
```
egracer, line 78: Any reason that the Semantic UI React convention wasn't used here?

https://react.semantic-ui.com/elements/input/#variations-right-left-labeled

client/src/components/NumberQuestion.test.js
```diff
@@ -1,59 +1,74 @@
...
-      <NumberQuestion questionKey={questionKey} />
-    </Provider>
+  let mockUi = renderWith(
+    <NumberQuestion questionKey={questionKey} {...renderOptions} />,
+    {
+      state: state,
+    }
   );
-  return { store: store, ...mockUi };
+  return { state: state, ...mockUi };
```
egracer, line 21: Is this ever used? I think it's redundant since you're returning the store from `renderWith`.

client/src/components/NumberQuestion.test.js
```diff
@@ -124,10 +139,9 @@ describe("NumberQuestion", () => {
     });
 
     it("has no a11y violations in error state", async () => {
-      const { getByLabelText, container } = mockUi;
```
egracer, line 127: In some of these tests, you call e.g. `screen.getByLabelText`, and in others you retrieve the `getByLabelText` function from `mockUi`. Could you standardize on one approach for all tests? I believe the convention for `renderWith` is to use `screen` unless pulling something like `container` or `store` out of the mock.

marcja: +1

client/src/util/testing.js
```diff
@@ -85,7 +85,7 @@ function renderWith(
     children: PropTypes.element.isRequired,
   };
 
-  return render(html, { wrapper: wrapper, ...renderOptions });
+  return { store, ...render(html, { wrapper: wrapper, ...renderOptions }) };
```
marcja, line 88: This is confusing. Don't do this. renderWith should still return what render returns. If you need a reference to the store, create one and pass it in to renderWith.

client/src/components/NumberQuestion.test.js
```diff
@@ -1,59 +1,74 @@
...
+        },
+      });
+
+      expect(container.getElementsByClassName("dollar sign")).toHaveLength(1);
+    });
+
+    it("renders with currency and is tab accessible", async () => {
+      const { container } = setupUI("HOUSEHOLD_SIZE", {
+        renderOptions: {
+          questionType: NumberQuestion.QuestionTypes.CURRENCY,
```
marcja, line 49: This is not a valid option to https://testing-library.com/docs/react-testing-library/api#render-options.

client/src/components/NumberQuestion.test.js
marcja: See below. I don't agree that renderWith's return type should be changed.

client/src/components/NumberQuestion.test.js
```diff
@@ -1,59 +1,74 @@
...
+    }
   );
-  return { store: store, ...mockUi };
+  return { state: state, ...mockUi };
 }
 
 describe("NumberQuestion", () => {
   let mockUi = null;
 
+  describe("conditional formatting", () => {
```
marcja, line 27: With the describe/it text you've got, it reads haltingly in test output. For example:

> NumberQuestion > conditional formatting > renders with explicit default answer type

What about:

> currency question > with input hint > renders
> currency question > with input hint > has input of type "number"
> currency question > with input hint > is required
> currency question > with input hint > has icon
> currency question > with input hint > has hint
> currency question > with input hint > has no a11y violations

client/src/components/NumberQuestion.test.js
```diff
@@ -1,59 +1,74 @@
...
+    it("renders with explicit default answer type", async () => {
+      const { container } = setupUI("HOUSEHOLD_SIZE", {
+        renderOptions: { questionType: NumberQuestion.QuestionTypes.DEFAULT },
+      });
+
+      expect(container.getElementsByClassName("dollar sign")).toHaveLength(0);
```
marcja, line 33: You don't need "container" here. Just `expect(screen.getBy*)`. 

Don't use getElementsByClassName; prefer the matchers in jest-dom. See https://testing-library.com/docs/guide-which-query

client/src/components/NumberQuestion.test.js
```diff
@@ -1,59 +1,74 @@
...
-}));
-
-const mockStore = configureMockStore();
-
-// TODO: Use renderWith.js
-function setupUI(questionKey, mockResponses) {
-  const store = mockStore({
-    responses: mockResponses || {},
-  });
+function setupUI(questionKey, { mockResponses = {}, renderOptions = {} } = {}) {
```
marcja, line 10: I personally think this setupUI function is vestigial and no longer needed once you start using renderWith and describe blocks.

client/src/components/NumberQuestion.test.js
tariqajyusuf: Added back but the test fails because it requires a change to `renderWith` to pass in the custom translation.

client/src/components/NumberQuestion.jsx
tariqajyusuf: Mainly out of not knowing, changed.

client/src/components/NumberQuestion.test.js
tariqajyusuf: Done!

client/src/components/NumberQuestion.test.js
tariqajyusuf: Commented out the line with a TODO since this PR is already getting a bit long.

client/src/components/NumberQuestion.jsx
```diff
@@ -9,12 +9,19 @@ import {
...
+  CURRENCY: "CURRENCY",
+};
+
 /**
  * @param {object} props
  *   - questionKey: a unique identifier for the question to be rendered
  *   - showErrorForNoEntry: a boolean value indicating whether to display
  *     an error prompt if the question has not been answered. Defaults
  *     to false. Will be true once the user attempts to submit the form.
+ *   - questionType: Determines how we format the number either as a "default" or
```
kebrahim, line 23: s/default/number

client/src/components/NumberQuestion.test.js
```diff
@@ -1,71 +1,130 @@
...
-      if (key !== "catalog:HOUSEHOLD_INCOME.hint") {
-        return key;
-      } else {
-        return null;
-      }
-    },
-    i18n: {
-      exists: (key) => {
-        return key !== "catalog:IS_EMPLOYED.hint";
+// Mock out i18next, ensuring one boolean question does not have hint text
```
kebrahim, line 12: s/boolean/number

client/src/components/NumberQuestion.test.js
```diff
@@ -1,71 +1,130 @@
...
-    <Provider store={store}>
-      <NumberQuestion questionKey={questionKey} />
-    </Provider>
-  );
-  return { store: store, ...mockUi };
-}
+  describe("with input hint", () => {
+    beforeEach(() => {
+      const initialState = {
+        questions: mockQuestions,
```
kebrahim, line 37: now that i look back at NumberQuestion.jsx, it doesn't seem to read/write the questions from Redux, so i don't believe you need to initialize this state w/ mockQuestions

client/src/components/NumberQuestion.test.js
```diff
@@ -167,22 +262,27 @@ describe("NumberQuestion", () => {
...
+    it("hides hint label", async () => {
+      const { getByText, getByLabelText } = renderWith(
+        <NumberQuestion questionKey="HOUSEHOLD_SIZE" />,
+        {
+          store: store,
+        }
+      );
+
+      expect(getByText("HOUSEHOLD_SIZE.text")).toBeInTheDocument();
+      // TODO(tariq): This test is broken because of how we mock out i18n, someone should fix this.
```
kebrahim, line 274: FWIW, i think ethan got mocking out the i18n working. is it possible you could reuse what he did in BooleanQuestion.test.js? https://github.com/ny/maslow/blob/master/client/src/components/BooleanQuestion.test.js#L14-L29

client/src/components/NumberQuestion.test.js
tariqajyusuf: The functionality breaks without this so I do need it.

client/src/components/NumberQuestion.test.js
tariqajyusuf: I'm doing the same thing right now but the problem is that we need to be able to manually trigger between having the hint exist and not exist which I can't figure out how to work out with mocks as we have it right now. I can try to tackle it here but it will take a long time, require some restructuring, and feels like it's better addressed in another PR.

client/src/components/NumberQuestion.jsx
mscharfstein: s/"default"/"number"

client/src/components/Section.jsx
mscharfstein: ping on this comment -- does this work as expected? i would be surprised if it did

tariqajyusuf: Yes it does, a slight change as per your recommendation makes this work. Added a test for this.

PR #101, opened 2020-07-22
Initialize Dropdown language to current language

### LINKED ISSUE #64: LanguageSwitcher should initialize value of Dropdown to i18next.language
Currently the LanguageSwitcher does not initialize the value of the Dropdown (that is, its value is unselected). This should be initialized with the value of [i18next.language](https://www.i18next.com/overview/api#language), which is the current language (either set automatically through [i18next language detection](https://github.com/i18next/i18next-browser-languageDetector) or set via [i18next.changeLanguage](https://www.i18next.com/overview/api#changelanguage)).

itssumertime:
Fixes #64. LanguageSwitcher now initializes Dropdown value to i18next.language. Also includes a small update to the header line color (to match NYS defined colors).


client/src/components/LanguageSwitcher.jsx
```diff
@@ -40,7 +49,7 @@ function LanguageSwitcher() {
         className="dropdown"
         icon={<Image src="globe.svg" floated="left" />}
```
kebrahim, line 50: is there a reason you removed the alt text for this image? we're getting an a11y warning for it

itssumertime: Oops, probably when switching to use Image.

client/src/components/LanguageSwitcher.jsx
```diff
@@ -30,17 +30,25 @@ function LanguageSwitcher() {
     { key: "zh", text: "中文", value: "zh" },
   ];
 
-  const changeLanguage = (e, { value }) => {
-    i18n.changeLanguage(value);
+  const getLanguageName = (languageCode) => {
+    for (const [index, value] of languageOptions.entries()) {
```
itssumertime, line 34: I get unused value "index" here but when I remove it, the fix I did stops working. What's the right way to do a React for loop where I just want to loop over all the entries and pull out the key/text from them?

kebrahim: languageOptions.values() work?

client/src/components/LanguageSwitcher.jsx
```diff
@@ -30,17 +30,25 @@ function LanguageSwitcher() {
...
+  const changeLanguage = (e, { language }) => {
+    i18n.changeLanguage(language);
   };
 
   return (
     <Grid>
       <Dropdown
         className="dropdown"
-        icon={<Image src="globe.svg" floated="left" />}
+        icon={<Image src="globe.svg" alt={t("globe-icon")} floated="left" />}
```
kebrahim, line 49: thanks. mind also adding one to the NYS logo?

kebrahim, line 49: should "globe-icon" appear in https://github.com/ny/maslow/blob/master/client/public/locales/en/translation.json ?

itssumertime: Just added it for globe and nyslogo. I'm going to say we can start with just the EN file until we decide how we want to translate icons.

client/src/components/LanguageSwitcher.jsx
```diff
@@ -30,17 +30,25 @@ function LanguageSwitcher() {
...
 
-  const changeLanguage = (e, { value }) => {
-    i18n.changeLanguage(value);
+  const getLanguageName = (languageCode) => {
+    for (const option of languageOptions) {
+      if (option.key === languageCode) {
+        return option.text;
+      }
+    }
+  };
```
marcja, line 39: This is fine, but you can also do:
```js
const getLanguageName = (languageCode) => 
  languageOptions.find((o) => o.key === "ko").text;
```

client/src/components/LanguageSwitcher.jsx
```diff
@@ -30,17 +30,25 @@ function LanguageSwitcher() {
     { key: "zh", text: "中文", value: "zh" },
   ];
```
marcja, line 31: Shouldn't these translations be in our translation files?

client/src/components/LanguageSwitcher.jsx
marcja: If you added the translations above into i18next, then you could just do: 

```js
t("language", { lng: "en" });
```

client/src/components/LanguageSwitcher.jsx
```diff
@@ -30,17 +30,25 @@ function LanguageSwitcher() {
...
 
   return (
     <Grid>
       <Dropdown
         className="dropdown"
-        icon={<Image src="globe.svg" floated="left" />}
+        icon={<Image src="globe.svg" alt={t("icon.globe")} floated="left" />}
         onChange={changeLanguage}
-        placeholder={t("Select Language")}
+        placeholder={getLanguageName(i18n.language)}
```
marcja, line 51: Where does placeholder even show up? Should you be using `text` instead? https://react.semantic-ui.com/modules/dropdown/#types-search-dropdown

Placeholder isn't very well supported for accessibility.

client/public/locales/en/translation.json
```diff
@@ -6,6 +6,10 @@
   "form": {
     "title": "Form"
   },
+  "icon": {
+    "globe": "Globe icon",
```
marcja, line 10: Should the text be "Select language" instead? I'm not clear how a person using a screenreader makes use of "Globe icon".

itssumertime: No because this is referring to just the icon. I added a separate one for Select Language.

client/src/components/LanguageSwitcher.jsx
itssumertime: This is the list of languages in the given language. I didn't think we had plan to support it saying "Spanish" and "Chinese" when you're set to English but saying "Espanol" etc if you switch the language.

Also the Dropdown takes in a list of "options" formatted this way.

client/src/components/LanguageSwitcher.jsx
itssumertime: I was using "placeholder" back when it said "Select a language" before any of them were clicked (shows up as light gray text). Now that I clarified we want current language set as the default, you're right that "text" should be used. Updated it.

client/src/components/LanguageSwitcher.jsx
itssumertime: I was getting the error "TypeError: Cannot read property 'text' of undefined" when I tried to do your suggestion but replace "ko" with the actual "languageCode" variable I was trying to pass in. Trying to figure out why but leaving for now.

kebrahim: add TODO to move these to i18next

client/public/locales/en/translation.json
marcja: Reading https://webaim.org/techniques/alttext/, it actually sounds like alt text should be omitted in this case, since the globe icon neither provides content or function. It previously did provide function when there was no other instruction to "Select language", but since you added that, this alt text is unnecessary and distracting to screenreaders.

client/public/locales/en/translation.json
```diff
@@ -29,6 +32,10 @@
     "page": "Step {{page}} of {{size}}",
     "title": "Services"
   },
+  "icon": {
+    "globe": "Globe icon",
+    "nyslogo": "New York State logo"
```
marcja, line 37: According to https://webaim.org/techniques/alttext/#logos, this should just be "New York State".

client/src/components/LanguageSwitcher.jsx
```diff
@@ -38,11 +46,11 @@ function LanguageSwitcher() {
     <Grid>
       <Dropdown
         className="dropdown"
-        icon={<Image src="globe.svg" floated="left" />}
+        icon={<Image src="globe.svg" alt={t("icon.globe")} floated="left" />}
```
marcja, line 49: Based on guidance in https://webaim.org/techniques/alttext/, this should just be `alt=""` now that you've added "Select language" as text.

client/src/components/LanguageSwitcher.jsx
itssumertime: I still don't understand why these would go into i18next. This list is meant for passing options into the Dropdown and unless we want it such that language names are all in English when the user has it set to English and are all in Korean when set to Korean (though I don't even know how you'd say "Bengali" in Korean...) then I don't see why any of this has to move to i18next. I'm not translating each one. I'm just mapping a language code to its full language name ***in that language***, not in ***all*** languages.

Issue #69, opened 2020-07-20
Implement visual styles to match NYS theme


mscharfstein:
Colors, fonts, spacing

kebrahim:
implement Semantic theme

kebrahim:
http://react.semantic-ui.com/theming

marcja:
Matching the NYS theme is definitely at risk for Sprint 1.

marcja:
Forked this into smaller issues:
- #150 
- #151 
- #152 
- #153 
- #154 

Closing this issue.

PR #102, opened 2020-07-22
support theming semantic-ui

marcja:
Works on #69 

This PR does not actually change any styles. It just rewires our project to support theming of semantic-ui. The theme is built from the /client/src/semantic-ui folder automatically as part of the build. Actually creating the theme will be a followup PR.


client/src/semantic-ui/site/collections/menu.overrides
```diff
@@ -0,0 +1,3 @@
+/*******************************
+         Site Overrides
+*******************************/
```
kebrahim, line 3: newline?

PR #103, opened 2020-07-22
Enable react-axe in development mode

kebrahim:
Works on #8 

This allows us to see a11y warnings directly in the debug console while navigating through the UI (see image below). I personally think this is useful, but LMK if you think it's too noisy.

TsFSXLROFVP


client/src/index.js
```diff
@@ -19,6 +22,11 @@ ReactDOM.render(
   document.getElementById("root")
 );
 
+if (process.env.NODE_ENV !== "production") {
```
marcja, line 25: Just curious -- can you test if the following works instead:

```js
if (__DEV__) {
```

Supposedly that constant is created by React and is preferred.

client/src/index.js
```diff
@@ -7,6 +7,9 @@ import { Provider } from "react-redux";
 import { BrowserRouter as Router } from "react-router-dom";
 import Routes from "./Routes.jsx";
 import * as serviceWorker from "./serviceWorker";
+import axe from "react-axe";
+
+const axeDebounceTimeoutMillis = 1000;
```
marcja, line 12: Move constant under your if(dev) check...

kebrahim: done

client/src/index.js
kebrahim: i can't seem to get react to recognize this even though it seems like it should be supported - https://reactjs.org/docs/codebase-overview.html#development-and-production

i'll keep investigating, but i'm going to submit this for now

PR #104, opened 2020-07-22
Added tab navigation test for BooleanQuestion

kebrahim:
Works on #8 #15 

Note that this didn't require any implementation changes, as it appears the BooleanQuestion, by itself, handles tab navigation correctly.


client/src/components/BooleanQuestion.test.js
```diff
@@ -88,15 +89,81 @@ describe("BooleanQuestion", () => {
...
+    it("supports tab navigation", () => {
+      const { getByLabelText } = mockUi;
+
+      const yesRadioButton = getByLabelText("radio.yes");
+      const noRadioButton = getByLabelText("radio.no");
+      expect(yesRadioButton).not.toBeChecked();
+      expect(noRadioButton).not.toBeChecked();
+
+      // Tab navigation cycles from body -> yes -> body without checking
+      expect(document.body).toHaveFocus();
```
marcja, line 102: Should the initial state instead be:

```js
expect(yesRadioButton).not.toHaveFocus();
expect(noRadioButton).not.toHaveFocus();
```

client/src/components/BooleanQuestion.test.js
```diff
@@ -88,15 +89,81 @@ describe("BooleanQuestion", () => {
...
+      userEvent.tab();
+      expect(yesRadioButton).toHaveFocus();
+      userEvent.tab();
+      expect(document.body).toHaveFocus();
+
+      expect(yesRadioButton).toBeChecked();
+      expect(noRadioButton).not.toBeChecked();
+    });
+  });
```
marcja, line 139: What about a test to change the response via keyboard? I assume that's tab, down arrow, spacebar?

client/src/components/BooleanQuestion.test.js
kebrahim: done

client/src/components/BooleanQuestion.test.js
kebrahim: looks like those keys aren't supported by user-event yet, but they're being actively worked on - https://github.com/testing-library/user-event/issues/354

PR #112, opened 2020-07-23
Fix lack of error on missing number and add tests to entire Form component 

### LINKED ISSUE #108: First page allows continuation without providing numeric input
I'm able to get past the first page of questions without passing in a numeric input for the first question.

Steps to repro (on Chrome, on Linux):
1) Add answers to all the boolean questions
2) Press 'continue'
3) Observe the validation error on the numeric question
4) Enter a valid number to remove the validation
5) Delete the valid number
6) Press 'continue'

Expected results:
- Should see validation telling me to enter a valid numeric input

Actual results:
- Navigated to the 2nd page

mscharfstein:
Fixes #108 

Works on tests for #13 . Remaining tests include: accessibility (incl tab test)


client/src/components/Form.test.js
```diff
@@ -3,6 +3,7 @@ import { fireEvent, screen } from "@testing-library/react";
 import React from "react";
 import Form from "./Form.jsx";
 import { renderWith } from "../util/testing";
+//import userEvent from "@testing-library/user-event";
```
itssumertime, line 6: delete line

client/src/components/Form.test.js
```diff
@@ -45,6 +46,12 @@ describe("form", () => {
       expect(back()).not.toBeInTheDocument();
       expect(done()).not.toBeInTheDocument();
     });
+
+    it("has FormEmpty heading", async () => {
```
itssumertime, line 50: So the FormEmpty heading AND FormIntro heading both exist? (Test below looks identical so just curious if both happen simultaneously.)

client/src/components/Form.test.js
```diff
@@ -123,5 +158,137 @@ describe("form", () => {
...
+    });
+
+    it("cannot move to next page because of empty responses", async () => {
+      const { init, next, back, done } = queryButtons(screen);
+      fireEvent.click(init());
+      fireEvent.click(next());
+
+      expect(back()).toBeInTheDocument();
+      expect(next()).toBeInTheDocument();
+      // if next had been allowed, we would expect done in the document
```
itssumertime, line 207: Isn't "next" allowed/expected? Comment doesn't make sense to me

client/src/components/Form.test.js
```diff
@@ -123,5 +158,137 @@ describe("form", () => {
...
+    });
+
+    it("cannot move to next page because of invalid response", async () => {
+      const { init, next, back, done } = queryButtons(screen);
+      fireEvent.click(init());
+      fireEvent.click(next());
+
+      expect(back()).toBeInTheDocument();
+      expect(next()).toBeInTheDocument();
+      // if next had been allowed, we would expect done in the document
```
itssumertime, line 249: same here

client/src/components/Form.test.js
```diff
@@ -123,5 +158,137 @@ describe("form", () => {
...
+      expect(done()).not.toBeInTheDocument();
```
kebrahim, line 250: mind also verifying that an error is displayed?

client/src/components/Form.test.js
```diff
@@ -123,5 +158,137 @@ describe("form", () => {
...
+          },
+          ADULTS_65_PLUS: {
+            type: "BOOLEAN",
+          },
+        },
+      };
+      renderWith(<Form />, { state: initialState });
+    });
+
+    it("cannot move to next page because of empty responses", async () => {
```
kebrahim, line 200: can you also add a test case for the bug you fixed? i.e. adding a numeric value, deleting it & then verifying you can't continue

client/package.json
```diff
@@ -80,6 +80,7 @@
     "react-redux": "^7.1.3",
     "react-router-dom": "^5.2.0",
     "react-scripts": "3.4.1",
+    "react-test-renderer": "^16.13.1",
```
kebrahim, line 83: where did you actually use this?

client/src/components/Form.test.js
mscharfstein: no, form empty is only if theres no form information (thats what diff about the tests)

client/src/components/Form.test.js
mscharfstein: no, next isn't allowed right now because the answers aren't valid (so if you press next the validation warnings pop up but dont take you to the next page)

client/src/components/Form.test.js
mscharfstein: ditto

client/src/components/Form.test.js
mscharfstein: i already test that on line 211, so i dont think it adds any benefit here (the only diff with this test is that the answer was "", not empty)

client/src/components/Form.test.js
mscharfstein: the test case is if the value is "" (as opposed to empty). that's what happens if you add a value and delete it - the key still exists in the object but the value is ""

client/package.json
mscharfstein: removed this

client/src/components/Form.test.js
itssumertime: Got it, that makes sense.

client/src/components/Form.test.js
```diff
@@ -290,5 +304,53 @@ describe("form", () => {
...
+      userEvent.tab();
+      expect(document.body).toHaveFocus();
+      expect(init()).not.toHaveFocus();
+
+      fireEvent.click(init());
+
+      // Tab navigation cycles from body -> through questions -> to continue/back
+      expect(document.body).toHaveFocus();
+      expect(next()).not.toHaveFocus();
+      userEvent.tab(); // onto first question
```
kebrahim, line 325: is it possible to verify that the first question now has focus? similar comment for the next two lines (second question, next button)

PR #113, opened 2020-07-23
Update FormSteps tests including a11y tests

### LINKED ISSUE #13: Render the QuestionForm

mscharfstein:
Fixes #13 (by finishing the form tests)

Includes a11y testing but no tab navigation because there's really nothing to click in the FormSteps component - the parent Form component has the navigation buttons and the children Section component has the questions


client/src/components/FormSteps.test.js
```diff
@@ -3,21 +3,110 @@ import { screen } from "@testing-library/react";
...
+test("form steps is accessible", async () => {
+  const form = [{ HOUSEHOLD_SECTION: ["HOUSEHOLD_SIZE", "ADULTS_65_PLUS"] }];
+  const { container } = renderWith(<FormSteps page={1} form={form} />);
+  expect(await axe(container)).toHaveNoViolations();
+});
+
+test("form steps is accessible with error", async () => {
+  const form = [{ HOUSEHOLD_SECTION: ["HOUSEHOLD_SIZE", "ADULTS_65_PLUS"] }];
+  const { container } = renderWith(
+    <FormSteps page={1} form={form} error={true} />
```
kebrahim, line 109: should this be showError instead of error?

client/src/components/FormSteps.test.js
```diff
@@ -3,21 +3,110 @@ import { screen } from "@testing-library/react";
...
+        id: "integer-question-HOUSEHOLD_SIZE",
+      });
+      expect(question1).toBeInTheDocument();
+      const numErrorText = screen.getByText("error.question.number");
+      expect(numErrorText).toBeInTheDocument();
+    });
   });
 });
+
+test("form steps is accessible", async () => {
```
kebrahim, line 100: could these tests be moved inside the `describe` sections above? it seems like the FormSteps used are the same

client/src/components/FormSteps.test.js
mscharfstein: yes thanks

client/src/components/FormSteps.test.js
mscharfstein: i dont think so because we need access to the container object

PR #114, opened 2020-07-23
Add Section tests

itssumertime:
This tests that sections have a title and questions.


client/src/components/Section.test.js
```diff
@@ -0,0 +1,78 @@
...
+import React from "react";
+import Section from "./Section.jsx";
+import { renderWith } from "../util/testing";
+
+expect.extend(toHaveNoViolations);
+
+describe("section", () => {
+  beforeEach(() => {
+    const initialState = {
+      form: [
```
kebrahim, line 13: you don't need to specify `form` or `responses` because they're not used by Section.jsx

client/src/components/Section.test.js
```diff
@@ -0,0 +1,78 @@
...
+
+    renderWith(
+      <Section sectionKey={sectionKey} questionKeys={questionKeys} />,
+      {
+        state: initialState,
+      }
+    );
+  });
+
+  it("section is accessible", async () => {
```
kebrahim, line 49: remove the "section" here. it'll inherit the "section" from the `describe` method above

client/src/components/Section.test.js
```diff
@@ -0,0 +1,78 @@
...
+    const { container } = renderWith(<Section />);
```
kebrahim, line 50: you don't need to call renderWith again, it was already called in the beforeEach method. you might just have to save it to a {container} above so you can access it here

client/src/components/Section.test.js
```diff
@@ -0,0 +1,78 @@
...
+    expect(await axe(container)).toHaveNoViolations();
+  });
+
+  it("section title and heading", async () => {
```
kebrahim, line 54: s/section/renders

client/src/components/Section.test.js
```diff
@@ -0,0 +1,78 @@
...
+    const section = screen.getByText("HOUSEHOLD_SECTION.title");
```
kebrahim, line 55: i don't think you need lines 55-56. you're already verifying the heading appears in lines 57-60

client/src/components/Section.test.js
```diff
@@ -0,0 +1,78 @@
...
+    expect(section).toBeInTheDocument();
+    const heading = screen.getByRole("heading", {
+      name: "HOUSEHOLD_SECTION.title",
+    });
+    expect(heading).toBeInTheDocument();
+  });
+
+  it("questions", async () => {
```
kebrahim, line 63: s/questions/renders questions

client/src/components/Section.test.js
```diff
@@ -0,0 +1,78 @@
...
+    const questionOne = screen.getByText("ADULTS_65_PLUS.text");
+    expect(questionOne).toBeInTheDocument();
+    const questionTwo = screen.getByText("HOUSEHOLD_SIZE.text");
+    expect(questionTwo).toBeInTheDocument();
+    const questionThree = screen.getByText("HOUSEHOLD_INCOME.text");
+    expect(questionThree).toBeInTheDocument();
+  });
+
+  it("radio buttons", async () => {
```
kebrahim, line 72: how about "renders radio buttons for boolean question"?

client/src/components/Section.test.js
kebrahim, line 49: mind also adding a keyboard-tab-navigation test that confirms you can navigate from question to question? you can see an example here: https://github.com/ny/maslow/blob/master/client/src/components/Form.test.js#L308

client/src/components/Section.test.js
itssumertime: I actually get "can't read the property of undefined" and also the sections are within "form". Even removing responses gives the same error, although I'm not using it, so I think it may be something with "renderWith" and Redux. It errors on the useSelector part of the Section.jsx file.

mscharfstein: yeah you need a least an empty responses (responses: {}) because NumberQuestion and BooleanQuestion will be called and those will try to look up a key in responses. you shouldnt need form i don't think

client/src/components/Section.test.js
mscharfstein: +1

client/src/components/Section.test.js
```diff
@@ -0,0 +1,78 @@
...
+    const radioYes = screen.getByRole("radio", { name: "radio.yes" });
+    expect(radioYes).toBeInTheDocument();
+    const radioNo = screen.getByRole("radio", { name: "radio.no" });
+    expect(radioNo).toBeInTheDocument();
+  });
+});
```
mscharfstein, line 78: can you add tests where the showErrorForNoEntry prop is true and theres just an empty response object so the answers are not valid: validate that the error text shows up

client/src/components/Section.test.js
itssumertime: Moved it to the bottom as a separate test.

client/src/components/Section.test.js
itssumertime: Done.

client/src/components/Section.test.js
itssumertime: Isn't this already tested in Boolean and NumberQuestion? The response object is empty in all of these tests anyway and error text doesn't show up.

kebrahim: yes, it's already tested in the questions, but we should also verify that the showError flag is correctly passed from Section to the questions

client/src/components/Section.test.js
```diff
@@ -69,10 +63,37 @@ describe("section", () => {
...
     expect(radioNo).toBeInTheDocument();
   });
+
+  it("supports tab navigation", () => {
+    // Tab navigation cycles from body -> through questions
+    expect(document.body).toHaveFocus();
+    userEvent.tab(); // onto first question
+    const question1 = screen.getByRole("spinbutton", {
+      name: "HOUSEHOLD_SIZE.text",
+      id: "integer-question-HOUSEHOLD_SIZE",
```
kebrahim, line 79: do you need to specify name, id & type? i think you might be able to only specify one of them (maybe just id or name)?

client/src/components/Section.test.js
```diff
@@ -69,10 +63,37 @@ describe("section", () => {
...
+      type: "number",
+    });
+    expect(question1).toHaveFocus();
+    userEvent.tab(); // onto second question
+    const question2 = screen.getByLabelText("radio.yes");
+    expect(question2).toHaveFocus();
+    userEvent.tab(); // onto third question
+    const question3 = screen.getByRole("spinbutton", {
+      name: "HOUSEHOLD_INCOME.text",
+      id: "integer-question-INCOME",
```
kebrahim, line 89: same comment about only needing to specify one key

client/src/components/Section.test.js
```diff
@@ -69,10 +63,37 @@ describe("section", () => {
...
+      type: "number",
+    });
+    expect(question3).toHaveFocus();
```
kebrahim, line 92: add one more `userEvent.tab()` & confirm that document.body has focus

client/src/components/Section.test.js
```diff
@@ -69,10 +63,37 @@ describe("section", () => {
...
+  });
+});
+
+test("has no accessibility violations", async () => {
+  const { container } = renderWith(<Section />);
```
kebrahim, line 97: mind also passing in the section & question keys? this test is just going to render an empty section

client/src/components/Section.test.js
marcja: just use name

client/src/components/Section.test.js
marcja: this seems brittle -- why add this?

itssumertime: Added for accessibility testing via tabs.

client/src/components/Section.test.js
itssumertime: Yes done.

client/src/components/Section.test.js
```diff
@@ -0,0 +1,99 @@
...
+        ADULTS_65_PLUS: {
+          type: "BOOLEAN",
+        },
+        HOUSEHOLD_INCOME: {
+          type: "NUMBER",
+        },
+      },
+    };
+    const sectionKey = "HOUSEHOLD_SECTION";
+    const questionKeys = [
```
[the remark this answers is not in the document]
itssumertime: It's not multiple sections (I thought so at first too since I used data from other tests). It's all the household section (adults 65+ is part of ages of household).

client/src/components/Section.test.js
```diff
@@ -0,0 +1,99 @@
...
+
+  it("renders heading", async () => {
+    const heading = screen.getByRole("heading", {
+      name: "HOUSEHOLD_SECTION.title",
+    });
+    expect(heading).toBeInTheDocument();
+  });
+
+  it("renders questions", async () => {
+    const questionOne = screen.getByText("ADULTS_65_PLUS.text");
```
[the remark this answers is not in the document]
itssumertime: Those don't work for the ones in div classes but I did use the roles for spinButtons.

The error for switching this to getByLabelText is: "Found a label with the text of: ADULTS_65_PLUS.text, however no form control was found associated to that label. Make sure you're using the "for" attribute or "aria-labelledby" attribute correctly." I think our formatting may be slightly different.

client/src/components/Section.test.js
itssumertime: Done.

PR #109, opened 2020-07-23
Accessibility for question components

egracer:
- Render BooleanQuestion radio buttons with semantic HTML to meet a11y guidelines and add CSS to style semantic HTML the same as other UI components. Add relevant `aria-*` tags.
- Add `aria-*` tags to NumberQuestion.
- Remove extraneous div nesting in the form.

PR #115, opened 2020-07-23
fix spacing and unnecessary markup

### LINKED ISSUE #111: Question groups appear too close together
When there is more than one question on a page, the questions appear too close together. I see this on both Chrome & Firefox on Linux:

BXr5R1MUSO9

marcja:
Fix #111 (among other things)

This PR has a lot more than CSS. Fixing the spacing required a two-pronged approach: 

1. **removing unnecessary markup:** if you look at the rendered markup after applying this PR, there are a LOT fewer `div` elements, which should also have the nice side effect of making the page more accessible.
2. **tweaking CSS:** the CSS in this PR gets us in the neighborhood -- it looks presentable now, but it could certainly benefit from more iteration. 

A few things to note:

- all CSS has been moved into CSS modules, using local classes wherever possible
- all CSS has been defined in em units, wherever possible -- this allows the app to be responsive as text size changes
- I also took care of some unrelated theming issues such as squaring off the corners of buttons and inputs

**NOTE:** unrelated to my changes, I noticed that there is a warning that appears when running tests:

```
    console.error node_modules/react-dom/cjs/react-dom.development.js:88
      Warning: A component is changing an uncontrolled input of type number to be controlled. Input elements should not switch from uncontrolled to controlled (or vice versa). Decide between using a controlled or uncontrolled input element for the lifetime of the component. More info: https://fb.me/react-controlled-components
          in input (created by Input)
          in div (created by Input)
          in Input (at NumberQuestion.jsx:59)
          in div (created by FormField)
          in FormField (at NumberQuestion.jsx:51)
          in NumberQuestion (at NumberQuestion.test.js:38)
          in Provider (at NumberQuestion.test.js:37)
```

I know what causes it (that is, initializing the value property of `<Input>` with a null/undefined value), but the fix is outside the scope of this PR.

Here are some screenshots:

page1
page2


client/src/components/NumberQuestion.module.css
```diff
@@ -0,0 +1,11 @@
+/* stylelint-disable selector-pseudo-class-no-unknown */
+
+:global .ui.form .grouped.fields .booleanQuestionRadioButton {
```
egracer, line 3: Is this for NumberQuestion? I think this modifies the boolean question style.

client/src/css/questions.css
```diff
@@ -1,9 +0,0 @@
-.questionHint {
```
egracer, line 1: I made some important changes in the questions.css file in #109 (had to migrate styles from Semantic components because they didn't use `fieldset` or `legend` tags and the `as` prop didn't work as expected). PTAL and let me know if there's anything I can to do help you sort through it.

client/src/components/NumberQuestion.module.css
kebrahim, line 3: numberQuestion?

mscharfstein: +1

marcja: Good catch. Will move.

marcja: done

marcja: done -- this got placed into the wrong module

client/src/css/questions.css
marcja: Can we chat ASAP this morning about this?

PR #124, opened 2020-07-24
Restore & refactor a11y code

### LINKED ISSUE #15: Render boolean questions

egracer:
A revert of #120 (with modifications), which was a revert of #109. Some notable differences from #109:

Fixes #15 

- Removed CustomRadioButton, found a simpler way to accomplish the same goal with CSS & aria-* attributes.
- Resolve React error about NumberQuestion being a controlled component by adding an empty string as a default value when the user has not provided a value.

Migrating `BooleanQuestion` tests to `renderWith` will be the next PR.


client/src/components/BooleanQuestion.jsx
```diff
@@ -18,56 +18,73 @@ import "./BooleanQuestion.module.css";
...
 
   return (
-    <Form.Field required error={showErrorForNoEntry && !answered}>
-      <label>{t(`catalog:${questionKey}.text`)}</label>
+    <Form.Field
+      required
+      error={showErrorMessage}
+      role="radiogroup"
+      aria-labelledby={a11yIds.questionTitle}
+      aria-describedby={`${a11yIds.hint} ${a11yIds.error}`}
```
kebrahim, line 51: this was actually causing an axe error after you had merged this PR the first time. can you check your local server to see if the error still appears? i was seeing it with the `axe` chrome extension, i believe - https://chrome.google.com/webstore/detail/axe-web-accessibility-tes/lhdoppojpmngadmnindnejefpokejbdd?hl=en-US

i believe the issue was that for fields where the hint wasn't being returned from the server, the <p> tag below w/ the a11yIds.hint id wasn't being rendered. so axe was throwing an error saying this was referencing a field that didn't exist.

egracer: Thanks, fixed!

client/src/components/BooleanQuestion.module.css
```diff
@@ -2,6 +2,17 @@
...
 }
+
+:global .ui.form .booleanQuestionRadioButton label {
+  margin: 0;
+  padding: 10px;
+  font-weight: 100;
+  cursor: pointer;
+}
+
+[type="radio"] {
```
marcja, line 15: You probably need `:global` to make this work. Are you sure this style is applied by the browser as is?

client/src/components/BooleanQuestion.module.css
```diff
@@ -2,6 +2,17 @@
 
 :global .ui.form .grouped.fields .booleanQuestionRadioButton {
   margin-top: 10px;
-  padding: 10px;
   border: 1px solid lightgray;
 }
+
+:global .ui.form .booleanQuestionRadioButton label {
+  margin: 0;
+  padding: 10px;
```
marcja, line 10: As a general rule, define sizes in em units, not pixels.

client/src/components/BooleanQuestion.test.js
```diff
@@ -175,12 +193,40 @@ describe("BooleanQuestion", () => {
...
   });
 
-  it("renders hint text if it exists", () => {
-    const { getByText, getByLabelText } = setup("ADULTS_65_PLUS");
+  describe("displays its error state", () => {
+    beforeEach(() => {
+      mockUi = setupWithError("IS_EMPLOYED");
+    });
+
+    it("when the question is not answered", () => {
```
marcja, line 201: The text in the `describe` and `it` seem reversed. The model is "given/when/then". So the describe block above should be "an unanswered question" and the it block should be "display its errors state".

client/src/components/BooleanQuestion.test.js
```diff
@@ -175,12 +193,40 @@ describe("BooleanQuestion", () => {
...
+      expect(await axe(container)).toHaveNoViolations();
+    });
+  });
+
+  describe("displays hint text", () => {
+    beforeEach(() => {
+      mockUi = setup("ADULTS_65_PLUS");
+    });
+
+    it("when hint text is defined", () => {
```
marcja, line 218: same here

client/src/components/BooleanQuestion.test.js
```diff
@@ -175,12 +193,40 @@ describe("BooleanQuestion", () => {
...
-    expect(getByText("catalog:ADULTS_65_PLUS.text")).toBeInTheDocument();
-    expect(getByLabelText("radio.yes")).toBeInTheDocument();
-    expect(getByLabelText("radio.no")).toBeInTheDocument();
+      expect(getByText("catalog:ADULTS_65_PLUS.hint")).toBeInTheDocument();
+      expect(getByText("catalog:ADULTS_65_PLUS.text")).toBeInTheDocument();
+      expect(getByLabelText("radio.yes")).toBeInTheDocument();
+      expect(getByLabelText("radio.no")).toBeInTheDocument();
+    });
+
+    it("with no a11y violations", async () => {
```
marcja, line 227: "has no a11y violations"

client/src/components/BooleanQuestion.module.css
egracer: It definitely works! Will add `:global` to be safe.

client/src/components/BooleanQuestion.test.js
egracer: Done

client/src/components/BooleanQuestion.test.js
egracer: Done

PR #129, opened 2020-07-24
Update i18next/i18next-parser to support ht, fallback

marcja:
There was an outstanding TODO to update i18next-parser to support the "ht" locale. It had been blocked on an [issue in i18next](https://github.com/i18next/i18next/issues/1491) (for which I submitted a [pull request](https://github.com/i18next/i18next/pull/1492)). That PR got merged and i18next released a new version.

This PR updates to the latest versions of i18next and i18next-parser and updates our i18next config to support "ht". I also ran i18next-parser (via `yarn i18n:extract`) to generated updated and synced translation files.

One notable configuration change: translations with the value "" will be considered not found and will use any configured fallback mechanism.


client/public/locales/bn/translation.json
```diff
@@ -10,8 +12,25 @@
     }
   },
   "form": {
+    "back": "",
```
itssumertime, line 15: Is the purpose of adding all the empty strings so that we know what keys are missing for that language?

Going forward should we be adding empty ones like this manually to keep track too, or is this some kind of yarn presubmit that notifies about doing so?

client/public/locales/en/translation.json
```diff
@@ -1,11 +1,6 @@
 {
-  "Language dropdown": "Language dropdown",
-  "No questions to display": "No questions to display",
```
itssumertime, line 3: Why are these being removed? I do think they should be something more "key-like" but I thought we wanted all strings to be in this file.

client/public/locales/en/translation.json
```diff
@@ -1,11 +1,6 @@
...
-  "Please select yes or no": "Please select yes or no.",
-  "Select Language": "Select Language",
   "button": {
-    "back": "Back",
-    "submit": "Submit"
```
itssumertime, line 8: Why is Submit removed?

client/public/locales/es/translation.json
```diff
@@ -1,7 +1,9 @@
 {
   "button": {
-    "back": "Atrás",
-    "submit": "Enviar"
```
itssumertime, line 4: Why is Submit removed? For this and all languages I used the Microsoft translation tool.

client/public/locales/es/translation.json
```diff
@@ -25,6 +28,9 @@
     "page": "Paso {{page}} de {{size}}",
     "title": "Servicios"
   },
+  "icon": {
+    "nyslogo": ""
```
itssumertime, line 32: Can we just put "New York State" since it's a name? Would the plan be to internationalize icon names for accessibility?

client/public/locales/ht/translation.json
```diff
@@ -1,7 +1,9 @@
 {
   "button": {
-    "back": "Retounen",
-    "submit": "Soumèt"
```
itssumertime, line 4: ditto on submit here and all files below

client/public/locales/bn/translation.json
marcja: Yes, the purpose of the empty strings is so that i18next knows that the string is missing and will use its fallback mechanism.

There is no reason to manually add these. `yarn i18n:extract` does it automatically when you run it.

client/public/locales/en/translation.json
marcja: They removed because they're no longer being referenced. `yarn i18n:extract` removed them because they weren't found in the codebase.

client/public/locales/en/translation.json
marcja: Same as above. It was not found in the codebase.

client/public/locales/es/translation.json
marcja: Same as above.

client/public/locales/es/translation.json
marcja: Leave that to the translator. With a blank entry, i18next will use English as its fallback (showing "New York State").

client/public/locales/ht/translation.json
marcja: Ditto here.

client/public/locales/en/translation.json
itssumertime: Ah okay I just walked through the app and it seems like it got switched to "Done" at some point.

PR #135, opened 2020-07-27
update stylelint to support css modules and usage rules

marcja:
Depends on #134 

Updates our stylelint configuration (which lints CSS) to support CSS modules and apply a few additional rules. Most importantly, absolute units (such as "px") are no longer allowed (except on border properties).


.vscode/extensions.json
```diff
@@ -3,6 +3,7 @@
     "dbaeumer.vscode-eslint",
     "esbenp.prettier-vscode",
     "msjsdiag.debugger-for-chrome",
+    "orta.vscode-jest",
```
kebrahim, line 6: need to merge in your other PR :)

marcja: Should be fine. This was branched from the previous PR.

PR #138, opened 2020-07-27
Migrate BooleanQuestion tests to use `renderWith` test harness

egracer:
Completes testing for #15. Simply a conversion of existing tests to use the `renderWith` harness for code consistency. No functional changes in the tests.


client/src/components/BooleanQuestion.test.js
```diff
@@ -88,22 +71,14 @@ describe("BooleanQuestion", () => {
       });
     });
 
-    it("displays an error if the question is not answered", () => {
```
tariqajyusuf, line 91: Is there a reason you're removing this test?

egracer: Yep, realized it's a duplicate of L195. Oops!

client/src/components/BooleanQuestion.test.js
```diff
@@ -1,83 +1,66 @@
...
 
     it("initially renders both 'Yes' and 'No' options", () => {
-      const { getByText, getByLabelText, queryByText } = mockUi;
-
-      expect(getByText("catalog:IS_EMPLOYED.text")).toBeInTheDocument();
-      expect(getByLabelText("radio.yes")).toBeInTheDocument();
-      expect(getByLabelText("radio.no")).toBeInTheDocument();
-      expect(queryByText("catalog:IS_EMPLOYED.hint")).not.toBeInTheDocument();
-      expect(queryByText("error.question.boolean")).not.toBeInTheDocument();
+      expect(screen.getByText("IS_EMPLOYED.text")).toBeInTheDocument();
```
kebrahim, line 51: is it possible to use `screen.getByRole` here?

client/src/components/BooleanQuestion.test.js
```diff
@@ -1,83 +1,66 @@
...
 describe("BooleanQuestion", () => {
   let mockUi = null;
 
   describe("without a response", () => {
+    let store = null;
+
     beforeEach(() => {
-      mockUi = setup("IS_EMPLOYED");
+      const initialState = {
+        questions: mockQuestions,
```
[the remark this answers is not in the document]
egracer: Done, I actually don't need to initialize questions because of the i18n mocking.

client/src/components/BooleanQuestion.test.js
egracer: Yes, done.

client/src/components/BooleanQuestion.test.js
```diff
@@ -1,83 +1,66 @@
...
+        },
       },
-    },
-  }),
-}));
+    }),
+  };
+});
 
 const mockStore = configureMockStore();
```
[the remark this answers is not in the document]
egracer: renderWith doesn't expose the store that it creates (which is required for some of these tests). So I need to configure a mock store and pass it in to renderWith. I moved this line closer to where that happens so the correlation is more clear.

PR #161, opened 2020-07-28
update catalog API to include rank of services

### LINKED ISSUE #160: Update catalog API to return ranked sequence of services
The priority of recommended services is static and unrelated to user input. Therefore we will have the API return a "rank" field that is an array of sectionKeys in priority order.

marcja:
Fix #160


server/src/main/java/gov/ny/its/hs/maslow/model/Catalog.java
```diff
@@ -135,6 +138,21 @@ public void setQuestions(Map<String, Question> questions) {
...
+   */
+  public List<String> getRank() {
+    return rank;
+  }
+
+  /**
+   * Sets the value for {@link #getRank()}
+   */
+  public void setRank(List<String> rank) {
+    this.rank = rank;
```
itssumertime, line 153: How is this setting the rank? Shouldn't it be returning indices corresponding to the keys (eg: "DOL_PUA" would return "0" since it's first in the list)? Just wondering how this rank is to be used.

marcja: It's an array. It has a fixed order.

PR #165, opened 2020-07-28
add DTOs for catalog endpoint, elide resources

### LINKED ISSUE #52: Remove unused translation resources from /api/explore/v1/catalog
Now that catalog translations are exposed dynamically through `/api/explore/v1/locales`, the translations should be removed from `/api/explore/v1/catalog`.  

Will introduce DTOs and ModelMapper (or equivalent) to translate our model objects to wire objects that don't have the translations.

marcja:
Fix #52 

Adds DTOs for the `/api/explore/v1/catalog` endpoint. This allows us to have a different schema on the wire than within the server. In this case, the DTO simply removes the resources (that is, the translation strings), since they are accessing via a different `/api/explore/v1/locales/en/catalog.json` endpoint.

Note that the "sections" does not appear in the DTO. As it turns out, this object _only_ had resources, so there was nothing to return.


server/src/main/java/gov/ny/its/hs/maslow/dto/CatalogDto.java
```diff
@@ -0,0 +1,146 @@
+package gov.ny.its.hs.maslow.dto;
+
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+/**
+ * Defines the catalog of questions and services that drive the business logic of application.
```
kebrahim, line 9: s/application/the application

server/src/main/java/gov/ny/its/hs/maslow/dto/CatalogDto.java
```diff
@@ -0,0 +1,146 @@
...
+
+  private List<String> rank = new ArrayList<>();
+
+  private Map<String, ServiceDto> services = new HashMap<>();
+
+  /**
+   * Gets the structure definition of the form. A form is comprised of one or more sections where
+   * each section has one or more questions. The form structure is defined by reference, not by
+   * value. That is, it is defined with section and question keys, which are in turn used to lookup
+   * section and question data from {@link #getSections()} and {@link #getQuestions()} respectively.
```
kebrahim, line 24: getSections doesn't actually exist anymore. i'm a little unclear why it doesn't. is that because it only contains strings? also, if it isn't referenced in this class, why do we need a SectionDto class at all?

server/src/main/java/gov/ny/its/hs/maslow/dto/CatalogDto.java
```diff
@@ -0,0 +1,146 @@
...
+   *     "CHILDREN_06_12",
+   *     "CHILDREN_00_05"
+   *  ]},{
+   *   "SECTION_HOUSEHOLD_INCOME": [
+   *     "HOUSEHOLD_INCOME"
+   *   ]},{
+   *   "SECTION_EMPLOYMENT_STATUS": [
+   *     "IS_EMPLOYED"
+   *   ]},{
+   *    "SECTION_ADDITIONAL_CRITERIA": [
```
kebrahim, line 60: indentation is off

server/src/main/java/gov/ny/its/hs/maslow/dto/ServiceDto.java
```diff
@@ -0,0 +1,69 @@
...
+
+  /**
+   * Gets the application URL for this service.
+   */
+  public String getApplicationUrl() {
+    return applicationUrl;
+  }
+
+  /**
+   * Sets the value of {@link #getApplyUrl(String)}
```
kebrahim, line 50: s/Apply/Application

server/src/main/java/gov/ny/its/hs/maslow/dto/CatalogDto.java
marcja: Please read the PR description where I had noted this.

As for SectionDto, you're right. It should be removed as it's not referenced.

server/src/main/java/gov/ny/its/hs/maslow/dto/CatalogDto.java
marcja: fixed

server/src/main/java/gov/ny/its/hs/maslow/dto/CatalogDto.java
marcja: SectionDto removed

server/src/main/java/gov/ny/its/hs/maslow/dto/CatalogDto.java
marcja: fixed

server/src/main/java/gov/ny/its/hs/maslow/dto/ServiceDto.java
marcja: fixed

server/src/main/java/gov/ny/its/hs/maslow/dto/CatalogDto.java
kebrahim: wait sorry, shouldn't this reference to getSections be removed, since the method doesn't exist?

marcja: Well, the method doesn't exist, but the underlying concept still does. The section key still needs to reference a valid section (as it appears in the application.yaml, for example). But given that this constraints is validated on the model, not the DTO, the comment can be removed (obviously, in a follow-up PR).

PR #167, opened 2020-07-28
remove references to getSections in CatalogDto comments

### LINKED ISSUE #52: Remove unused translation resources from /api/explore/v1/catalog

marcja:
Fix #52


server/src/main/java/gov/ny/its/hs/maslow/dto/CatalogDto.java
```diff
@@ -63,11 +63,6 @@
    *     "IS_STUDENT"
    *   ]}
    * ]
-   *
-   * Each string, such as "SECTION_HOUSEHOLD" or "HOUSEHOLD_SIZE", is a key into its respective
-   * dictionary ({@link #getSections()} and {@link #getQuestions()}, respectively.)
```
kebrahim, line 68: should we remove the reference to `getSections()` on line 24 too?

marcja: done

Issue #146, opened 2020-07-27
Create Service Card UI


egracer:
Each card will have:

- Category (e.g. Food)
- Service Title (e.g. SNAP)
- Service Description (a few sentences at most)
- Additional Criteria (only collapsed if description + criteria size > 256 pixels)
- ‘Add to report’ button which adds service to “n selected” frame. ‘Add to report’ button text changes to ‘Added to report’. When ‘added to report’ button is pressed, service is removed from selected list.

[See the mocks](https://preview.uxpin.com/642233d8d1d76a804e0ad41477dc05e51ccf21fc#/pages/130713745/comment/sitemap,comments?mode=i).

PR #168, opened 2020-07-28
Create Report & Rank slices

### LINKED ISSUE #158: Update Redux to support services selection UI
We'll need to update Redux to support the functionality of the selection page. This includes:

- [x] Storing the array of service keys that represents the ranking of the services (includes selector, updater)
- [x] Storing the array of selected service keys to be included in the report (includes selector, updater)
- [x] (Maybe) Adding a selector to servicesSlice to retrieve the list of services that do not have a formula (to be displayed in the "Additional Services" bucket)

Depends on #157, #160

egracer:
Fixes #158.

The `report` slice is very similar to the `responses` slice -- it just keeps track of which services will be added to the report.
The `rank` slice saves the static ranking of services from #160 into redux.

I also added a convenience selector for the `services` slice to split services by those that do and do not contain a formula.

This will also unblock the remainder of #146.


client/src/App.test.js
```diff
@@ -62,6 +62,8 @@ describe("app with mock store", () => {
     expect(store.getActions()[2].payload.questions).toEqual(mockQuestions);
     expect(store.getActions()[3].type).toEqual("services/updateServices");
     expect(store.getActions()[3].payload.services).toEqual(mockServices);
+    expect(store.getActions()[4].type).toEqual("rank/updateRank");
+    expect(store.getActions()[4].payload.services).toEqual(mockServices);
```
kebrahim, line 66: this should be checking `payload.rank`

client/src/features/services/servicesSlice.test.js
```diff
@@ -1,38 +1,109 @@
...
-});
+describe("The services slice", () => {
+  it("has an empty initial state", () => {
+    expect(reduceServices(undefined, { type: undefined })).toEqual({});
+  });
+
+  it("updates services", () => {
+    const services = {
+      DOL_UI: {
+        enabled: true,
```
kebrahim, line 15: this is probably a TODO, but i assume we should be filtering by enabled services, as well, right? we should never display a disabled service in the report.

client/src/features/report/reportSlice.js
```diff
@@ -0,0 +1,22 @@
+import { createSlice } from "@reduxjs/toolkit";
+
+export const reportSlice = createSlice({
```
marcja, line 3: Why isn't this just part of the view slice?

client/src/App.js
```diff
@@ -26,6 +27,7 @@ function App() {
       dispatch(updateSections(data));
       dispatch(updateQuestions(data));
       dispatch(updateServices(data));
+      dispatch(updateRank(data));
```
marcja, line 30: This is one of the several reasons I think we are creating too many slices. All of these slices are backed by the exact same data and are updated only once at the exact same time. These really should be just one slice with more selectors.

That said, consider maybe just having a single read-only slice for the whole services area and reuse the view slice (that is, add more actions) to store view content like selected services? If you did this, a lot of the code below would drop out (since many of your actions and reducers would no longer be needed).

client/src/App.test.js
egracer: Done.

client/src/features/services/servicesSlice.test.js
egracer: Agreed -- shouldn't this happen on the backend though?

client/src/features/rank/rankSlice.test.js
```diff
@@ -0,0 +1,27 @@
...
+
+  it("updates the provided ranking", () => {
+    const rank = ["DOL_PUA", "DOL_UA"];
+    const newState = reduceRank([], {
+      type: updateRank.type,
+      payload: {
+        rank: rank,
+      },
+    });
+    expect(newState).toEqual(rank);
```
itssumertime, line 17: How exactly is this updating the ranking? Shouldn't the test try giving an initial rank and then swapping the order of the rank to see if it ends up with the changed order?

client/src/features/report/reportSlice.js
```diff
@@ -0,0 +1,22 @@
...
+  name: "report",
+  initialState: {
+    // indicates if a service will be included in the report, e.g.
+    // DOL_PA
+    // DOL_UA
+    // but should start empty
+  },
+  reducers: {
+    updateReport: (state, action) => {
+      state[action.payload] = !state[action.payload];
```
marcja, line 13: Can you elaborate on what's going on here? Did you mean to rename "updateReport" to "toggleReport"? The code here is a bit opaque.

client/src/features/services/servicesSlice.test.js
```diff
@@ -1,38 +1,109 @@
...
-  expect(reduceServices(undefined, { type: undefined })).toEqual({});
-});
+describe("The services slice", () => {
+  it("has an empty initial state", () => {
+    expect(reduceServices(undefined, { type: undefined })).toEqual({});
+  });
+
+  it("updates services", () => {
+    const services = {
+      DOL_UI: {
```
itssumertime, line 14: You use "DOL_UI" and "DOL_UA" in some places - make consistent?

client/src/features/rank/rankSlice.test.js
egracer: "Updating" the ranking is similar to "updating" the order of questions on the form. It's all determined by the server. The UI just stores it -- at no point can the user change the ranking of services (just which services they want to keep in their report).

client/src/features/services/servicesSlice.test.js
egracer: Nice catch, fixed.

client/src/features/services/servicesSlice.js
```diff
@@ -14,4 +14,18 @@ export const { updateServices } = servicesSlice.actions;
 
 export const selectServices = (state) => state.services;
 
+export const selectServicesByFormula = (state) => {
```
marcja, line 17: Why do this (what is effectively a) groupBy? Not clear how this is intended to be used. Maybe remove this chunk from this PR and include it when you can show how it will be used in context.

client/src/features/rank/rankSlice.test.js
itssumertime: Oh got it, that makes sense then. Thanks!

client/src/features/services/servicesSlice.test.js
marcja: It should, but it is not currently. Added #170

client/src/features/services/servicesSlice.js
egracer: The intention is for it to be used by the Services page to split the services into the appropriate buckets. I added it in this PR since it's loosely related to pulling info out of Redux and it was a request in the issue description. Plus, Sumer is going to use it nearly immediately. Happy to pass the buck to @itssumertime if it's critical that we only add code when it's used in the same PR.

client/src/App.js
egracer: Let's sync on this tomorrow since I'm about to head out -- I'm kinda following but need a bit more information. Also want to make sure I'm not diving down a rabbit hole rewriting our entire app's redux structure, which I think would be a separate task if we decide we want to simplify.

client/src/features/report/reportSlice.js
marcja: I think it's unnecessary to split them, and it's leading to a lot of extra code -- but it's not _wrong_.

client/src/features/report/reportSlice.js
marcja: I think `toggleReport` would be much clearer. Or even better, `toggleIncludeInReport` or similar.

client/src/features/services/servicesSlice.js
marcja: I harbor a concern that this may not be the right design, but I can't substantively comment on it since I can see how it's being used. I'd say include this particular chunk in the PR where it's being used.

client/src/App.js
marcja: Fair. That's why I suggested doing something different just for the _new_ code, not yet refactoring the old code. Let's chat tomorrow.

client/src/features/report/reportSlice.js
egracer: Done.

client/src/features/report/reportSlice.js
egracer: Discussed offline, will leave this in as an intermediate step pending a team decision on #173.

client/src/App.js
egracer: Discussed offline, going to integrate rank into the services slice to reduce boilerplate and simplify the process of storing the catalog to Redux.

client/src/features/services/servicesSlice.js
egracer: Removed.

PR #172, opened 2020-07-28
filter services to only those enabled

### LINKED ISSUE #170: API should not return services where enabled = false
This field is not currently checked.

marcja:
Fix #170


server/src/main/java/gov/ny/its/hs/maslow/controller/ExploreController.java
```diff
@@ -28,7 +29,12 @@
    */
   @PostConstruct
   public void init() {
-    this.catalogDto = new ModelMapper().map(catalog, CatalogDto.class);
+    catalogDto = new ModelMapper().map(catalog, CatalogDto.class);
+
+    // we only want this endpoint to return services that are enabled
+    catalogDto.setServices(
+      Maps.filterEntries(catalogDto.getServices(), service -> service.getValue().getEnabled())
```
kebrahim, line 36: if possible, please add a unit test for this

Issue #163, opened 2020-07-28
Create eligibility parser logic


mscharfstein:
- Each service has a formula field to determine possible eligibility, with Excel-like formatting
          - Examples: “=IS_DISABLED”, “=HOUSEHOLD_INCOME < HOUSEHOLD_SIZE * 1000”, “=IFS(IS_STUDENT, HOUSEHOLD_INCOME < 1000, IS_MILITARY, HOUSEHOLD_INCOME < 2000, TRUE, FALSE)”

- All supported functions are defined in Maslow design
         - Logical/Operator functions are P0s
                - Logical: AND, FALSE, IF, IFS, NOT, OR, TRUE, XOR
                - Operator: ADD (+), DIVIDE (/), EQ (=), GT (>), GTE (>=), LT (<), LTE (<=), MINUS (-), MULTIPLY (*), NE (<>), POW (^), UMINUS (-)
         - Math/Statistical functions are P1s
                 - Math: ABS, CEILING, FLOOR, INT, MOD, POWER, PRODUCT, QUOTIENT, ROUND, SIGN, SQRT, SUM
                 - Statistical: MAX, MIN

- Create Tokens to represent the pieces of the equations, and a parser/actions to act on different parts of the formulas (using Chevrotain library) and 

- Interpret formula based on responses from user (must assign values to Identifiers based on responses from users)

mscharfstein:
For reference, formulas that will need to be supported:
HOUSEHOLD_INCOME <= (1345 + HOUSEHOLD_SIZE * 737)
NOT(IS_EMPLOYED)
AND(MONTHLY_HOUSEHOLD-INCOME*12 <= 16140 + HOUSEHOLD_SIZE * 8840, OR(CHILDREN_06_12, CHILDREN_00_05), OR(IS_EMPLOYED, IS_STUDENT))
OR(CHILDREN_13_17, CHILDREN_06_12, CHILDREN_00_05)
OR(IS_PREGNANT, CHILDREN_13_17, CHILDREN_06_12, CHILDREN_00_05)
IS_MILITARY
OR(CHILDREN_00_05, IS_PREGNANT)
IS_STUDENT
AND(IS_STUDENT, HOUSEHOLD_INCOME * 12 < 50550)
AND(IS_MILITARY, IS_DISABLED)
IS_DISABLED
CHILDREN_00_05
AND(IS_MILITARY, IS_STUDENT)
AND(IS_MILITARY, IS_DISABLED)
[...]

PR #177, opened 2020-07-29
Add support for Equals operator

mscharfstein:
Works on #163 

Adds in support for Equals operator as a Subclass of Relational Operators. Note that this handles the case where there is an Equals relational expression (not an '=' sign at the very beginning of the equation representing the Sheets formatting; that will need to be handled in a separate PR).

Note that relational operators have the lowest precedence (https://www.mathworks.com/help/matlab/matlab_prog/operator-precedence.html), below all of the numerical/mathematical operators (e.g. 1+2+3=6 * 1 .. 1+2+3 and 6*1 should be evaluated prior to comparing the two with the '=') so relational operators are the first rule defined now. I believe that AND, OR, etc will be even lower precedence, so that rule will need to come before the relational expression when it is added (@tariqajyusuf FYI)

Also adjusts the exports from the different files to more closely match the examples where the three files are separate


client/src/services/formulas/actions.js
```diff
@@ -21,7 +24,27 @@ class EvaluatorInterpreter extends BaseCstVisitor {
 
   expression(ctx) {
     // visiting an array is equivalent to visiting its first element.
-    return this.visit(ctx.additionExpression);
+    return this.visit(ctx.relationalExpression);
+  }
+
+  relationalExpression(ctx) {
+    let lhsValue = this.visit(ctx.lhs);
+    let result = lhsValue; // result starts out as the value of the LHS; might be overridden by a boolean value if there is a relationalOperator
```
marcja, line 32: Line too long. Move comment above?

client/src/services/formulas/actions.js
```diff
@@ -21,7 +24,27 @@ class EvaluatorInterpreter extends BaseCstVisitor {
...
+
+    // "rhs" key may be undefined as the grammar defines it as optional (MANY === zero or more).
+    if (ctx.rhs) {
+      ctx.rhs.forEach((rhsOperand, idx) => {
```
marcja, line 36: Does this imply that a formula like "=TRUE = TRUE = FALSE" is valid and equals FALSE?

client/src/services/formulas/actions.js
```diff
@@ -88,6 +111,10 @@ class EvaluatorInterpreter extends BaseCstVisitor {
     return Math.pow(base, exponent);
   }
 
+  initialEqualsExpression(ctx) {
```
marcja, line 114: startFormulaExpression?

client/src/services/formulas/actions.js
mscharfstein: done

client/src/services/formulas/actions.js
mscharfstein: didn't mean to include this in this PR. removed

client/src/services/formulas/actions.js
mscharfstein: yes, it would imply that. in the parser, the relationalExpression is expressed as having a lhs and then MANY (which means 0 or more) groupings of relationalOperator + rhs. we need the definition to allow for 0 of those groupings, and i don't think there's a way to specify 0 or 1 of them (vs. 0 or more). if we explicitly only want having one relational operator to be valid, we should look into that more. otherwise, this works as you stated and i added a test to confirm

mscharfstein: note that right now, TRUE/FALSE are not handled as special strings and strings aren't handled in general

client/src/services/formulas/evaluator.test.js
```diff
@@ -20,11 +20,27 @@ describe("Evaluations", () => {
...
   it("can evaluate an expression with power function", () => {
     expect(evaluate("1 + power(2,2)").value).toBe(5);
   });
+
+  it("can evaluate an expression with an Equals operator", () => {
+    expect(evaluate("1 = 2").value).toBe(false);
+  });
+
+  it("can evaluate an expression with multiple Equals operators", () => {
+    expect(evaluate("1=1=2").value).toBe(false);
```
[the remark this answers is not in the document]
mscharfstein: yeah :) will be able to make a test that makes more sense once we have the other operators

Issue #141, opened 2020-07-27
Top-level Services UI


itssumertime:
Outer layer Services UI with two buckets:
1) Services that match the formula
2) All other services (that don't match formula - which is a link to the main page ny.gov/services)

Need to also support the empty state - if no services match, then show empty state. Need to double check this part though since we may want to just show bucket 2, the link to all services.

kebrahim:
just a note that "all other services" and "link to ny.gov/services" are actually in separate "sections" on the page. this mock shows it best (https://preview.uxpin.com/642233d8d1d76a804e0ad41477dc05e51ccf21fc#/pages/130996614/comment/sitemap?mode=i)

Services that match are in Bucket 1:
![4vWqAu8pXOo](https://user-images.githubusercontent.com/2208244/88597841-25d8d480-d036-11ea-869f-04bd64968efc.png)

Services that don't match are in Bucket 2, then the text w/ the link is below that:
![oiA4UXbh0eJ](https://user-images.githubusercontent.com/2208244/88597871-31c49680-d036-11ea-80e8-617f3ca2d6b7.png)

Issue #139, opened 2020-07-27
Service Buckets UI


itssumertime:
- Bucket title (eg: "Here are 3 services you may qualify for")
- List of service cards (eg: "Memorial Scholarship", "SNAP", "HEAP")
- Support ranking - update catalog API to pass back an array of sorted service keys, then display the service keys in the ranked order (eg: "SNAP" first since it has highest potential benefit, and then the other services)

PR #179, opened 2020-07-29
Add Buckets and basic Services page layout

itssumertime:
Addresses #139 and some of #141. 

Adds Buckets of ServiceCards (dependent on Ethan's code, just including here to unblock myself) with the given serviceKeys. 
The Services page contains two Buckets - one with the services the user may qualify for and one with the additional services. It also contains a "see all services" section which we will later have to add a link in.
Adjusted divider CSS to distinguish the overall yellow header divider from the ones going in between Buckets.
The serviceKeys are currently hardcoded in the Services class, with a TODO to read them in from Redux/filtering logic once that code is available for use.
Adds i18n translations for English and Spanish keys and adds empty keys for other languages to keep track of what is left to translate.


client/src/components/AppHeader.jsx
```diff
@@ -21,7 +21,7 @@ function AppHeader() {
         src={logo}
       />
       <LanguageSwitcher />
-      <Divider />
+      <Divider className={styles.headerDivider} />
```
egracer, line 24: Nice didn't realize you could do this 👍

client/src/components/Bucket.jsx
```diff
@@ -0,0 +1,33 @@
...
+
+/**
+ * @param {object} props
+ *   - serviceKeys: an array of objects that holds service keys
+ *     Example: serviceKeys = ["DOL_PUA", "SNAP", "MemorialScholarship"];
+ * @returns {React.ReactElement} This will return a list of service cards.
+ */
+function Bucket({ serviceKeys }) {
+  const serviceCards = [];
+  for (let key of serviceKeys) {
```
egracer, line 14: s/`let`/`const`

client/src/components/Bucket.jsx
```diff
@@ -0,0 +1,33 @@
...
+    serviceCards.push(<ServiceCard serviceKey={key} />);
+  }
+
+  return (
+    <Container fluid textAlign="center">
```
egracer, line 19: Why center aligned? The mocks look left aligned.

client/src/components/Bucket.test.js
```diff
@@ -0,0 +1,28 @@
...
+
+expect.extend(toHaveNoViolations);
+
+describe("bucket", () => {
+  beforeEach(() => {
+    const serviceKeys = ["DOL_PUA", "DOL_UA", "SNAP"];
+    renderWith(<Bucket serviceKeys={serviceKeys} />);
+  });
+
+  it("renders service cards", async () => {
```
egracer, line 16: Could you write a separate test or expand this one to make sure that 3 separate cards are rendered? It looks to me like this test only checks that the first card renders a button, which feels more in scope for the ServiceCard component than the Bucket component.

client/src/components/Services.jsx
```diff
@@ -0,0 +1,32 @@
...
+ * match the formula and a bucket of all other services.
+ */
+function Services() {
+  // TODO: read service keys from Redux/filtering logic
+  // TODO: add link for ny.gov/services.
+
+  const { t } = useTranslation();
+
+  return (
+    <Container fluid textAlign="center">
```
egracer, line 18: Same comment re: center aligning.

client/src/components/ServiceCard.jsx
```diff
@@ -0,0 +1,54 @@
+import PropTypes from "prop-types";
```
egracer, line 1: Does this code need to merged, instead of using a placeholder Semantic UI `Card`? I think it'll be easier to review the ServiceCard submission in one go, rather than piecemeal across 2 PRs.

client/src/components/Bucket.test.js
```diff
@@ -0,0 +1,28 @@
...
+    const button = screen.getAllByRole("button", {
+      name: "BUTTON.ADDTOREPORT",
+    })[0];
+    expect(button).toBeEnabled();
+  });
+});
+
+test("has no axe violations", async () => {
+  const { container } = renderWith(<Bucket serviceKeys={["SNAP"]} />);
```
kebrahim, line 25: you can put this test inside of the describe, using something similar to what we did in AppHeader.test.js - https://github.com/ny/maslow/blob/master/client/src/components/AppHeader.test.js#L11-L15

client/src/components/Bucket.test.js
```diff
@@ -0,0 +1,28 @@
...
+  beforeEach(() => {
+    const serviceKeys = ["DOL_PUA", "DOL_UA", "SNAP"];
+    renderWith(<Bucket serviceKeys={serviceKeys} />);
+  });
+
+  it("renders service cards", async () => {
+    const button = screen.getAllByRole("button", {
+      name: "BUTTON.ADDTOREPORT",
+    })[0];
+    expect(button).toBeEnabled();
```
kebrahim, line 20: is it possible to also verify there are 3 cards rendered?

client/src/components/Services.jsx
```diff
@@ -0,0 +1,32 @@
...
+      <h1 className={styles.font}> {t("services.mayQualify")} </h1>
```
kebrahim, line 19: i'm wondering if the bucket should display its own title (as opposed to putting that rendering here)? i don't feel super strongly about it, but i guess that's how i envisioned this. any particular reason why you put it here instead?

client/src/components/Services.jsx
```diff
@@ -0,0 +1,32 @@
...
+import { Container, Divider } from "semantic-ui-react";
+import Bucket from "./Bucket.jsx";
+import styles from "./Services.module.css";
+
+/**
+ * @returns {React.ReactElement} This will contain a bucket of services that
+ * match the formula and a bucket of all other services.
+ */
+function Services() {
+  // TODO: read service keys from Redux/filtering logic
```
kebrahim, line 12: add TODO to handle empty state when no services match

client/src/components/Services.test.js
```diff
@@ -0,0 +1,41 @@
...
+
+  it("renders view all services", async () => {
+    const heading = screen.getByRole("heading", {
+      name: "services.viewAll",
+    });
+    expect(heading).toBeInTheDocument();
+  });
+});
+
+test("has no axe violations", async () => {
```
kebrahim, line 37: similar comment about moving this test inside of `describe`

client/src/components/Services.test.js
```diff
@@ -0,0 +1,41 @@
...
+describe("services", () => {
+  beforeEach(() => {
+    renderWith(<Services />);
+  });
+
+  it("renders services", async () => {
+    const heading = screen.getByRole("heading", {
+      name: "services.mayQualify",
+    });
+    expect(heading).toBeInTheDocument();
```
kebrahim, line 19: mind also verifying the service cards show up, similar to how we did it in FormSteps.test.js for questions - https://github.com/ny/maslow/blob/master/client/src/components/FormSteps.test.js

client/public/locales/en/translation.json
```diff
@@ -35,6 +35,13 @@
     "no": "No",
     "yes": "Yes"
   },
+  "services": {
+    "additional": "Additional Services",
+    "mayQualify": "We found these services that you may qualify for",
+    "offer": "Here are some other services that we offer.",
```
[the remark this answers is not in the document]
itssumertime: This text is to match this mock: https://preview.uxpin.com/642233d8d1d76a804e0ad41477dc05e51ccf21fc#/pages/130713745/specification/sitemap?mode=i

I'll add the piece about no matching services in a separate PR.

client/src/components/Bucket.jsx
itssumertime: Thanks for pointing this out, just fixed it.

client/src/components/Bucket.test.js
itssumertime: Yeah I made it 3 buttons for now since I didn't know how to check if a ServiceCard is rendered but I'm going to work on that now.

client/src/components/ServiceCard.jsx
itssumertime: Planning to just wait for your code to be submitted (added comment to the top).

client/public/locales/en/translation.json
```diff
@@ -35,6 +35,13 @@
     "no": "No",
     "yes": "Yes"
   },
+  "services": {
+    "additional": "Additional Services",
+    "mayQualify": "We found these services that you may qualify for",
```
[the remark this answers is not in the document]
itssumertime: Added to JSX code since comments can't go in the json.

client/public/locales/en/translation.json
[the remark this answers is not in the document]
itssumertime: Added TODO in jsx file.

client/src/components/Bucket.test.js
itssumertime: Done, thanks for the link.

client/src/components/Bucket.test.js
itssumertime: Verified via 3 buttons but going to work on figuring out if there's a better way.

client/src/components/Services.jsx
itssumertime: To be honest, I just hadn't thought about it that hard but you're right that makes so much more sense and is cleaner so I've moved it.

client/src/components/AppHeader.module.css
```diff
@@ -16,6 +10,12 @@
   margin-right: 1em;
 }
 
+.headerDivider {
```
marcja, line 13: I think it's supposed to be kebab-case in the CSS and camelCase in the JS. `.header-divider` should magically map to `styles.headerDivider` in the code.

client/src/components/Bucket.jsx
```diff
@@ -0,0 +1,42 @@
...
+  const serviceCards = [];
+  for (const key of serviceKeys) {
+    serviceCards.push(<ServiceCard serviceKey={key} />);
+  }
+
+  return (
+    <Container fluid>
+      <h1 className={styles.font}> {title} </h1>
+      <b> {additionalText} </b>
+      {serviceCards}
```
marcja, line 25: What if this array ends up empty (for example, when serviceKeys = [])?

client/src/components/Bucket.jsx
```diff
@@ -0,0 +1,42 @@
...
+  );
+}
+
+Bucket.propTypes = {
+  title: PropTypes.string,
+  additionalText: PropTypes.string,
+  serviceKeys: PropTypes.array,
+};
+
+Bucket.defaultProps = {
```
marcja, line 36: Are you sure you want default values for all of these properties defined? Should none of them be required?

client/src/components/Bucket.jsx
```diff
@@ -0,0 +1,42 @@
...
+  return (
+    <Container fluid>
+      <h1 className={styles.font}> {title} </h1>
+      <b> {additionalText} </b>
+      {serviceCards}
+    </Container>
+  );
+}
+
+Bucket.propTypes = {
```
marcja, line 30: Should any of these properties be required?

client/src/components/Bucket.test.js
```diff
@@ -0,0 +1,61 @@
+import "@testing-library/jest-dom/extend-expect";
```
marcja, line 1: It turns out this import isn't needed. It's already being imported via `setupTests.js`.

client/src/components/Bucket.test.js
```diff
@@ -0,0 +1,61 @@
...
+
+  it("renders heading", async () => {
+    const heading = screen.getByRole("heading", {
+      name: "services.mayQualify",
+    });
+    expect(heading).toBeInTheDocument();
+  });
+
+  it("renders 3 service cards", async () => {
+    const serviceCard1 = screen.getByText("DOL_PUA.name");
```
marcja, line 34: Have you looked at what this looks like in the Accessibility pane of Chrome DevTools? Are you sure there isn't a role to use instead of getByText?

client/src/components/Bucket.test.js
```diff
@@ -0,0 +1,61 @@
...
+    const serviceCard3 = screen.getByText("SNAP.name");
+    expect(serviceCard3).toBeInTheDocument();
+  });
+
+  it("renders 3 buttons, one for each service card", async () => {
+    const buttonOne = screen.getAllByRole("button", {
+      name: "BUTTON.ADDTOREPORT",
+    })[0];
+    expect(buttonOne).toBeEnabled();
+    const buttonTwo = screen.getAllByRole("button", {
```
marcja, line 47: screen.getAllByRole returns an array. Why are you calling it multiple times? Why not:

```js
const [buttonOne, buttonTwo, buttonThree] = screen.getAllByRole("button");
```

client/src/components/Bucket.test.js
```diff
@@ -0,0 +1,61 @@
...
+    expect(serviceCard2).toBeInTheDocument();
+    const serviceCard3 = screen.getByText("SNAP.name");
+    expect(serviceCard3).toBeInTheDocument();
+  });
+
+  it("renders 3 buttons, one for each service card", async () => {
+    const buttonOne = screen.getAllByRole("button", {
+      name: "BUTTON.ADDTOREPORT",
+    })[0];
+    expect(buttonOne).toBeEnabled();
```
marcja, line 46: Is there any disabling behavior? How could this expectation ever be false?

client/src/components/ServiceCard.jsx
```diff
@@ -0,0 +1,55 @@
...
+          {t(`catalog:${serviceKey}.description`)}
+        </Card.Description>
+
+        {/** TODO: Add support for additional criteria */}
+      </Card.Content>
+      <Card.Content extra>
+        <Button
+          className="service-card-button"
+          basic={!addedToReport}
+          primary={addedToReport}
```
marcja, line 36: Does toggling the `primary` property impact accessibility? Or is it only an aesthetic property?

client/src/components/ServiceCard.jsx
```diff
@@ -0,0 +1,55 @@
...
+          {t(`catalog:${serviceKey}.name`)}
+        </Card.Header>
+        <Card.Description>
+          {t(`catalog:${serviceKey}.description`)}
+        </Card.Description>
+
+        {/** TODO: Add support for additional criteria */}
+      </Card.Content>
+      <Card.Content extra>
+        <Button
```
marcja, line 33: Should be using the [Toggle variation](https://react.semantic-ui.com/elements/button/#variations-toggle) in Semantic UI React instead.

client/src/components/ServiceCard.jsx
```diff
@@ -0,0 +1,55 @@
...
+          className="service-card-button"
+          basic={!addedToReport}
+          primary={addedToReport}
+          color="blue"
+          onClick={() => setAddedToReport(!addedToReport)}
+        >
+          {!addedToReport && t("button.addToReport").toUpperCase()}
```
marcja, line 40: Seems a little messy. Maybe extract the content logic to a function?

client/src/components/ServiceCard.jsx
```diff
@@ -0,0 +1,55 @@
...
+            </>
+          )}
+        </Button>
+      </Card.Content>
+    </Card>
+  );
+}
+
+ServiceCard.propTypes = {
+  serviceKey: PropTypes.string,
```
marcja, line 54: Required?

client/src/components/ServiceCard.module.css
```diff
@@ -0,0 +1,17 @@
+/* stylelint-disable selector-pseudo-class-no-unknown */
+
+:global .ui.fluid.card,
+:global .ui.label {
+  border-radius: 0;
```
marcja, line 5: This CSS selector is _global_, meaning it's trying to set it across the entire app. Did you mean only to apply this to ServiceCard?

client/src/components/ServiceCard.module.css
```diff
@@ -0,0 +1,17 @@
...
+}
+
+:global .ui.card > .content > .header.header-margin {
+  margin-top: 0.5em;
+}
+
+:global .ui.button.service-card-button {
```
marcja, line 12: Can you get this to work with just a `.service-card-button` selector rather than making it global?

client/src/components/Services.jsx
```diff
@@ -0,0 +1,41 @@
...
+  // TODO: handle empty state when no services match
+  // TODO: read service keys from Redux/filtering logic
+  // TODO: add link for ny.gov/services.
+  // TODO: update translation.json with number of services in
+  // "services.mayQualify" key.
+
+  const { t } = useTranslation();
+
+  return (
+    <Container fluid>
```
marcja, line 21: Do you need `fluid` here?

client/src/components/Services.module.css
```diff
@@ -0,0 +1,8 @@
+.bucketDivider {
```
marcja, line 1: kebab-case

client/src/components/Services.module.css
```diff
@@ -0,0 +1,8 @@
...
+  background-color: #949494;
+  border: unset;
+}
+
+.font {
```
marcja, line 6: Is there a more semantic name for this style? Such as "service-header"?

client/src/components/Services.test.js
```diff
@@ -0,0 +1,50 @@
...
+
+  it("renders view all services", async () => {
+    const heading = screen.getByRole("heading", {
+      name: "services.viewAll",
+    });
+    expect(heading).toBeInTheDocument();
+  });
+
+  it("renders service cards", async () => {
+    const serviceCard1 = screen.getByText("DOL_PUA.name");
```
marcja, line 38: Is there an alternative getByRole or getByLabelText that works?

client/src/components/Services.module.css
itssumertime: Changing it means the divider color isn't respected but I'm not sure why. Leaving for now and maybe it can be addressed in the overall CSS cleanup that will eventually happen for formatting.

client/src/components/ServiceCard.jsx
egracer: Oy this makes for some very difficult CSS changes. It *really* wants that toggle button to be green 😂. Will follow up with you offline.

client/src/components/ServiceCard.jsx
egracer: Purely aesthetic, though if we use the toggle variation potentially becomes unnecessary (assuming we can figure out the CSS magic).

client/src/components/ServiceCard.jsx
egracer: Done

client/src/components/ServiceCard.jsx
egracer: Yes, done

client/src/components/ServiceCard.module.css
egracer: No, for these I intended it to be set across the entire app. Seems like we've settled on a consistency of no border radius.

client/src/components/ServiceCard.module.css
egracer: I have tried but without success. Seems like a CSS precedence issue that might be solvable but I'm out of my depth. Will sync with you tomorrow.

client/src/components/Services.jsx
egracer: Yes, it allows the card to [fill the ancestor's width](https://react.semantic-ui.com/elements/container/#types-container). Doesn't look great without it. Could probably achieve the same thing with manual CSS but couldn't think of a reason we wouldn't want the behavior.

client/src/components/AppHeader.module.css
marcja: Sorry, I was wrong. https://github.com/css-modules/css-modules#naming

client/src/components/Bucket.jsx
marcja: I'm saying, what is the expected user experience supposed to be when it's empty? "We found these services that you may qualify for"... and blankness?

client/src/components/Bucket.jsx
marcja: https://reactjs.org/docs/typechecking-with-proptypes.html

client/src/components/ServiceCard.module.css
marcja: But isn't it non-obvious that this global setting is buried in a CSS module that purports to be scoped to ServiceCard?

client/src/components/Bucket.jsx
itssumertime: Ah okay yeah I have a TODO in the Services.jsx to handle the empty state in a separate PR (once I'm actually reading from Redux).

client/src/components/Bucket.test.js
itssumertime: Changed it to "toBeInTheDocument".

client/src/components/Services.test.js
itssumertime: No. There are "headings" for the 3 sections and then "buttons" for each of the services I've added, but the buttons are not unique. These will have to be updated once the redux and filtering are added in follow up PRs anyway.

client/src/components/Bucket.jsx
marcja: Are you then saying that Bucket here _MUST_ receive a non-empty array of serviceKeys? That it is required _and_ its length must be > 0?

client/src/components/Bucket.jsx
marcja: Maybe an empty section shouldn't be a Bucket....

client/src/components/Bucket.test.js
```diff
@@ -0,0 +1,52 @@
...
+    const initialState = {
+      title: "services.mayQualify",
+      serviceKeys: ["DOL_PUA", "DOL_UA", "SNAP"],
+    };
+    ({ container } = renderWith(
+      <Bucket
+        title={initialState.title}
+        serviceKeys={initialState.serviceKeys}
+      />,
+      { state: initialState }
```
marcja, line 21: Stylistically, I find this code to be simpler as:

```js
const options = {
  state: {
    title: ...,
    serviceKeys: ...
  }
};

renderWith(<Bucket ... />, options);
```

client/src/components/Bucket.test.js
```diff
@@ -0,0 +1,52 @@
...
+    ));
+  });
+
+  it("renders heading", async () => {
+    const heading = screen.getByRole("heading", {
+      name: "services.mayQualify",
```
marcja, line 27: I think this isn't the ideal way to do it. getByRole returns an error if the target isn't found. So, in a sense, the expect below is meaningless (that is, if the call to getByRole succeeds, it _must_ be in the document). What about:

```js
const heading = screen.getByRole("heading");
expect(heading).toHaveTextContent("services.mayQualify");
```

itssumertime: Done.

client/src/components/Bucket.test.js
itssumertime: Yes, there are only the buttons for service cards and headings for buckets so I tested those.

client/src/components/Services.test.js
itssumertime: I did it with the headings of sections but will expand these once I read in real data.

client/src/components/Bucket.test.js
itssumertime: Updated it even though it looks pretty similar to me since the only "option" is "state".

client/src/components/Bucket.jsx
itssumertime: No, an empty array of service keys is currently supported.

client/src/components/Bucket.test.js
```diff
@@ -0,0 +1,46 @@
...
+  });
+
+  it("renders 3 buttons, one for each service card", async () => {
+    const [buttonOne, buttonTwo, buttonThree] = screen.getAllByRole("button");
+    expect(buttonOne).toBeInTheDocument();
+    expect(buttonTwo).toBeInTheDocument();
+    expect(buttonThree).toBeInTheDocument();
+  });
+
+  it("has no axe violations", async () => {
```
kebrahim, line 43: can be a TODO, but please also add a keyboard navigation test

client/src/components/Services.jsx
```diff
@@ -0,0 +1,40 @@
+import React from "react";
+import { useTranslation } from "react-i18next";
+import { Container, Divider } from "semantic-ui-react";
+import Bucket from "./Bucket.jsx";
+import styles from "./Services.module.css";
+
+/**
+ * @returns {React.ReactElement} This will contain a bucket of services that
+ * match the formula and a bucket of all other services.
```
kebrahim, line 9: it's actually a bucket of all services that don't have formulas

client/src/components/Services.test.js
```diff
@@ -0,0 +1,48 @@
...
+    expect(serviceCard1).toBeInTheDocument();
+    const serviceCard2 = screen.getByText("DOL_UA.name");
+    expect(serviceCard2).toBeInTheDocument();
+    const serviceCard3 = screen.getByText("SNAP.name");
+    expect(serviceCard3).toBeInTheDocument();
+  });
+
+  it("has no axe violations", async () => {
```
kebrahim, line 45: can be a TODO, but please add keyboard navigation test

itssumertime: Going to manually add an issue so as not to trigger all the TODO regeneration for now.

client/src/components/Bucket.test.js
itssumertime: https://github.com/ny/maslow/issues/239

client/src/components/Bucket.jsx
```diff
@@ -18,10 +18,15 @@ function Bucket({ title, additionalText, serviceKeys }) {
...
   }
 
+  // TODO: Handle empty serviceKeys case for Bucket.
+
   return (
     <Container fluid>
-      <h1 className={styles.bucketHeader}> {title} </h1>
-      <b> {additionalText} </b>
+      <Header as="h2" className={styles.bucketHeader}>
+        {" "}
```
marcja, line 26: What are these `{" "}` doing here?

client/src/components/Bucket.test.js
```diff
@@ -11,33 +11,26 @@ describe("bucket", () => {
...
-    expect(serviceCard2).toBeInTheDocument();
-    const serviceCard3 = screen.getByText("SNAP.name");
-    expect(serviceCard3).toBeInTheDocument();
+    const [heading1, heading2, heading3, heading4] = screen.getAllByRole(
+      "heading"
+    );
+    expect(heading1).toHaveTextContent("services.mayQualify");
+    expect(heading2).toHaveTextContent("DOL_PUA.name");
+    expect(heading3).toHaveTextContent("DOL_UA.name");
+    expect(heading4).toHaveTextContent("SNAP.name");
```
marcja, line 33: Nice!

PR #180, opened 2020-07-29
Add all of the variable identifiers to the parser

mscharfstein:
Works on #163 

Adds handling of the numerical and boolean identifiers. Throws error in actions if a provided variable is not one of the responses. Could do some refactoring and throw the error in parsing instead, if that is better


client/src/services/formulas/actions.js
```diff
@@ -17,6 +17,17 @@ const {
   LessThan,
   LessThanOrEqual,
   NotEqual,
+  IsEmployed,
```
marcja, line 20: Nope, you can't do this. These have to be dynamic. The questions aren't fixed, they are defined by data. This is much like implementing a parser for a programming language. You can't bake in the names of every conceivable variable name that a programmer could come up with.

mscharfstein: :)

tariqajyusuf: +1 (c.f. our chat just now)

client/src/services/formulas/actions.js
```diff
@@ -116,7 +117,18 @@ class EvaluatorInterpreter extends BaseCstVisitor {
...
     } else if (ctx.powerFunction) {
       return this.visit(ctx.powerFunction);
-    } // TODO(mscharfstein/tariq): Add support for variables/identifiers
+    } else if (ctx.Identifier) {
+      const variable = this.responses[ctx.Identifier[0].image];
+      if (variable) {
+        return variable;
+      } else {
+        throw Error(
+          "Errors detected!\n No response for provided variable '" +
```
marcja, line 126: useTranslation?

client/src/services/formulas/actions.js
```diff
@@ -136,13 +148,9 @@ class EvaluatorInterpreter extends BaseCstVisitor {
...
   // Add action for logical expression (handle ND, FALSE, IF, IFS, NOT, OR,
   // TRUE, XOR)
-  // Add action for Identifier (set variable like HOUSEHOLD_SIZE to its value)
 }
 
-// We only need a single interpreter instance because our interpreter has no state.
-const interpreter = new EvaluatorInterpreter();
-
-const evaluate = (text) => {
+const evaluate = (text, responses = {}) => {
```
marcja, line 153: s/text/formula/g

client/src/services/formulas/evaluator.test.js
```diff
@@ -67,4 +67,28 @@ describe("Evaluations", () => {
   it("can evaluate an expression with an NotEqual operator", () => {
     expect(evaluate("1 <> 2").value).toBe(true);
   });
+
+  it("can evaluate an expression with IS_EMPLOYED identifier", () => {
+    expect(evaluate("IS_EMPLOYED", { IS_EMPLOYED: true }).value).toBe(true);
```
marcja, line 72: Nice!!!

client/src/services/formulas/actions.js
marcja: Also, I think this error message presupposes too much knowledge about the app. Maybe: "Errors detected!\n Unknown variable '{{image}}'".

client/src/services/formulas/lexer.js
```diff
@@ -93,6 +93,11 @@ const NotEqual = createToken({
   categories: RelationalOperator,
 });
 
+const Identifier = createToken({
+  name: "Identifier",
+  pattern: /\w+/,
```
marcja, line 98: Regex should be: /[A-Z0-9_]{1,250}/. See https://support.google.com/docs/answer/63175.

mscharfstein: updated, but not sure if we want to be this strict. i know that's what sheets implements, but do we want to prevent admins from entering variables with no lower cases etc?

client/src/services/formulas/actions.js
mscharfstein: you can't use useTranslation hook outside of a React component. I think it's probably okay for the error to be thrown in english?

client/src/services/formulas/lexer.js
marcja: Yes. We should also be validating this on the server too.

Issue #142, opened 2020-07-27
Validate all service formulas


kebrahim:
Tasks include:

- [ ] Create a new hidden page (with a new route), which won't be user-facing
- [ ] Retrieve list of services from the API
- [ ] Run the service formula through the parser and show validation errors

PR #182, opened 2020-07-29
Added more services to application.yaml

kebrahim:
Works on #142 #141 #139 

Added 4 more services to application.yaml, including two which don't have any formula. So, they should both show up in the "Additional services" bucket.


server/src/main/resources/application.yaml
```diff
@@ -89,10 +89,32 @@ maslow:
...
     - DOL_UA
+    - VICTIM_COMP
+    - CHILDCARE
+    - EAA
+    - DOL_PUA
 
     services:
+      CHILDCARE:
+        enabled: true
+        formula: "=AND(MONTHLY_HOUSEHOLD-INCOME*12 <= 16140 + HOUSEHOLD_SIZE * 8840, OR(CHILDREN_06_12, CHILDREN_00_05), OR(IS_EMPLOYED, IS_STUDENT))"
```
marcja, line 102: `s/INCOME*12/INCOME * 12/g`

server/src/main/resources/application.yaml
```diff
@@ -89,10 +89,32 @@ maslow:
...
+        information-url: https://ocfs.ny.gov/main/childcare/subsidy_program.asp
+        resources:
+          en:
+            name: Child Care Subsidy Program
+            category: Family
+            description: |
+              Receive assistance paying for some or all of the cost of child care services.
+            eligibility: |
+              You must need child care to work, look for work, or attend employment training.
+            instructions: |
```
marcja, line 113: FYI ONLY: You only need the vertical bar if you are following the key with multiline text. This is also perfectly fine:
```
eligibility: You must need child care to work, look for work, or attend employment training.
instructions: To apply contact your local Department of Social Services
```

server/src/main/resources/application.yaml
```diff
@@ -153,3 +175,71 @@ maslow:
...
+        enabled: true
+        formula:
+        application-url: https://vsp.ovs.ny.gov/vsp/index.html
+        information-url: https://ovs.ny.gov/victim-compensation
+        resources:
+          en:
+            name: Victim Compensation
+            category: Crime
+            description: |
+              Receive assistance paying for – including but not limited to medical bills, counseling expenses, burial and funeral costs, and lost wages – resulting from being victims of crime.
```
marcja, line 229: Wrap at 80 chars

server/src/main/resources/application.yaml
kebrahim: done

server/src/main/resources/application.yaml
kebrahim: done

server/src/main/resources/application.yaml
kebrahim: so, i think this is actually causing problems if you combine the vertical bar with wrapping at 80 chars. it seems to be inserting newlines directly into the text, where they shouldn't be: 
cmfGgHGLpsj

i updated all of the text to how i *think* it should be formatted. this is the resulting JSON:
4ZgJqmWK1yB

server/src/main/resources/application.yaml
```diff
@@ -99,19 +99,16 @@ maslow:
...
             category: Family
-            description: |
-              Receive assistance paying for some or all of the cost of child care services.
-            eligibility: |
-              You must need child care to work, look for work, or attend employment training.
-            instructions: |
-              To apply contact your local Department of Social Services
+            description: Receive assistance paying for some or all of the cost of child care services.
+            eligibility: You must need child care to work, look for work, or attend employment training.
+            instructions: To apply contact your local Department of Social Services
```
marcja, line 111: Sorry, my comment before was annotated as "FYI ONLY". I didn't mean for you to make this change. Per our chat discussion, we should probably leave all markdown fields as literal text blocks (that is, using the vertical bar syntax).

kebrahim: done

PR #183, opened 2020-07-29
ServiceCard Component

### LINKED ISSUE #146: Create Service Card UI

egracer:
Fixes #146.

Add ServiceCard UI.

- [x] Saves whether it's included in the final report Redux
- [x] a11y (heading, button)
- [x] i18n (button text, service content)
- [x] tests

Remaining:

Some CSS issues with the `semantic-ui-react` toggle variation for button, which is stubbornly insisting on being green (rather than blue) when active.

service-card


client/src/components/ServiceCard.module.css
```diff
@@ -0,0 +1,36 @@
...
+  top: 2em;
+}
+
+:global .ui.card > .content > .ui.label {
+  position: relative;
+  top: -3em;
+  margin: 0;
+}
+
+:global .ui.button.service-card-button {
```
marcja, line 20: Try making it local, if you haven't already, and rename this selector to just ".serviceCardButton" and referencing it as {style.serviceCardButton} in the code.

client/src/components/ServiceCard.test.js
```diff
@@ -0,0 +1,91 @@
...
+  let mockUi = null;
+  let store = null;
+
+  describe("when the service is not selected", () => {
+    beforeEach(() => {
+      const initialState = {
+        report: {},
+      };
+      store = mockStore(initialState);
+      mockUi = renderWith(<ServiceCard serviceKey="DOL_UA" />, {
```
marcja, line 22: My personal preference is to simplify the code with, but it's not a formal standard:
```
const options = {
  store: mockStore({
    report: {}
  })
};
renderWith(<ServiceCard />, options);
```

marcja: Also, you only use the `container` field of mockUI. It would simplify the code in this file to only store `container` at the describe scope. Though you will need to use this funky syntax:

```js
({ container } = renderWith(...));
```

client/src/components/ServiceCard.test.js
```diff
@@ -0,0 +1,91 @@
...
+        store: store,
+      });
+    });
+
+    it("renders the 'Added To Report' button", () => {
+      const button = screen.getByRole("button");
+
+      expect(button).toHaveTextContent(/button.addedToReport/i);
+      expect(button).toHaveClass("active");
+    });
```
marcja, line 73: Please add a test for clicking on a services that's already been added to a report (that is, toggling it back again).

client/src/components/ServiceCard.module.css
```diff
@@ -0,0 +1,36 @@
...
+  padding: 1.25em 2em;
+  font-weight: bold;
+  font-size: 0.75em;
+  border-radius: 0;
+}
+
+/* Pixels used here per instructions at 
+https://webaim.org/techniques/css/invisiblecontent/ */
```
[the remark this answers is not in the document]
egracer: Good to know, changed.

client/src/components/ServiceCard.test.js
egracer: Agreed that would be more concise code, is there a way I can do that while still accessing the store from the subsequent tests? I need to check that an action has been saved to the store, which is why I was assigning it to the variable at a higher scope.

That container change is definitely funky! Done.

client/src/components/ServiceCard.test.js
egracer: The test for clicking in the button in the default state is on line 40, the test for clicking on it in the added state is on line 75. Can I name the tests differently to make that distinction clearer? I tried to call out those two cases with the describe blocks.

marcja: Nope, I think that I just didn't see it in the diff view. All good.

PR #185, opened 2020-07-30
Add boolean literals and logical functions

tariqajyusuf:
Works on #163 

We add a new `LogicalValue` to enable boolean literals, factor out `powerFunc` to a `function` to be able to easily add new function behaviors, add functionality and tests for logical operators.

I left some todos to do cleanups of tests and factor out functions to a dispatcher since this PR is pretty big.


client/src/services/formulas/lexer.js
```diff
@@ -93,6 +92,59 @@ const NotEqual = createToken({
   categories: RelationalOperator,
 });
 
+const BooleanLiteral = createToken({
+  name: "BooleanLiteral",
+  pattern: /(TRUE|FALSE)/,
+});
+
+// ------------------ FUNCTIONS ------------------
+const Function = createToken({
```
mscharfstein, line 101: Function isn't used anywhere is it?

client/src/services/formulas/evaluator.test.js
```diff
@@ -67,4 +68,44 @@ describe("Evaluations", () => {
   it("can evaluate an expression with an NotEqual operator", () => {
     expect(evaluate("1 <> 2").value).toBe(true);
   });
+
+  it("can evaluate a boolean literal", () => {
```
mscharfstein, line 72: can you add tests using AND and OR not just with TRUE and FALSE? but for example AND(1+2=3, HOUSEHOLD_SIZE) and chaining ... AND(1+2=3, OR(1=1, 1=2))

client/src/services/formulas/actions.js
```diff
@@ -114,23 +130,70 @@ class EvaluatorInterpreter extends BaseCstVisitor {
...
     } // TODO(mscharfstein/tariq): Add support for variables/identifiers
   }
 
+  // TODO(tariq): factor this out to a separate file to do function dispatching and make it much easier to update and add new behavior.
+  function(ctx) {
+    let params = [];
+    let functionName = ctx.FunctionName[0];
+    let result = null;
+
+    // "rhs" key may be undefined as the grammar defines it as optional (MANY === zero or more).
```
mscharfstein, line 144: update comment

client/src/services/formulas/evaluator.test.js
mscharfstein: to use HOUSEHOLD_SIZE, will have to get your LGTM and submit PR #180 first

client/src/services/formulas/actions.js
```diff
@@ -114,23 +130,70 @@ class EvaluatorInterpreter extends BaseCstVisitor {
...
+    if (ctx.param) {
+      ctx.param.forEach((param, idx) => {
+        params.push(this.visit(param));
+      });
+    }
+
+    // TODO(tariq): Check on param count
```
marcja, line 151: Meaning, return Error objects when the wrong number of arguments are provided?

client/src/services/formulas/actions.js
```diff
@@ -114,23 +130,70 @@ class EvaluatorInterpreter extends BaseCstVisitor {
...
+    if (tokenMatcher(functionName, PowerFunc)) {
+      const base = params[0];
+      const exponent = params[1];
+      result = Math.pow(base, exponent);
+    } else if (tokenMatcher(functionName, And)) {
+      result = params[0];
+      for (const param of params) {
```
marcja, line 158: I find this line needlessly different from the version in the other clauses. Have minor differences to very similar code creates bugs and adds complexity.

client/src/services/formulas/actions.js
```diff
@@ -114,23 +130,70 @@ class EvaluatorInterpreter extends BaseCstVisitor {
...
+      result = params[0];
+      for (let i = 1; i < params.length; i++) {
+        result = result || params[i];
+      }
+    } else if (tokenMatcher(functionName, Not)) {
+      result = !params[0];
+    } else if (tokenMatcher(functionName, Xor)) {
+      result = params[0];
+      for (let i = 1; i < params.length; i++) {
+        result = (result || params[i]) && !(result && params[i]);
```
marcja, line 171: The definition of this function is Sheets is "Returns true if an odd number of the provided arguments are logically true, and false if an even number of the arguments are logically true." See https://support.google.com/docs/answer/9116491.

client/src/services/formulas/actions.js
```diff
@@ -114,23 +130,70 @@ class EvaluatorInterpreter extends BaseCstVisitor {
...
+      } else {
+        result = params[2];
+      }
+    } else if (tokenMatcher(functionName, Ifs)) {
+      for (let i = 0; i < params.length; i += 2) {
+        if (params[i]) {
+          result = params[i + 1];
+          break;
+        }
+      }
```
marcja, line 185: Nice!

client/src/services/formulas/actions.js
```diff
@@ -114,23 +130,70 @@ class EvaluatorInterpreter extends BaseCstVisitor {
...
+
+    // "rhs" key may be undefined as the grammar defines it as optional (MANY === zero or more).
+    if (ctx.param) {
+      ctx.param.forEach((param, idx) => {
+        params.push(this.visit(param));
+      });
+    }
+
+    // TODO(tariq): Check on param count
+    if (tokenMatcher(functionName, PowerFunc)) {
```
marcja, line 152: These chained if/else-if statements seems like a [code smell](https://en.wikipedia.org/wiki/Code_smell) to me. Warrants some consideration about how to refactor if we intend to add a bunch more functions.

client/src/services/formulas/evaluator.test.js
```diff
@@ -67,4 +68,44 @@ describe("Evaluations", () => {
...
+    expect(evaluate("TRUE").value).toBe(true);
+    expect(evaluate("FALSE").value).toBe(false);
+  });
+
+  it("can evaluate AND statements", () => {
+    expect(evaluate("AND(TRUE,FALSE)").value).toBe(false);
+    expect(evaluate("AND(TRUE,TRUE)").value).toBe(true);
+    expect(evaluate("AND(TRUE,FALSE,TRUE)").value).toBe(false);
+    expect(evaluate("AND(TRUE,TRUE,TRUE)").value).toBe(true);
```
marcja, line 81: I mentioned this in our watercooler chat, but you can also reduce the code here a bit by using [describe.each](https://jestjs.io/docs/en/api#describeeachtablename-fn-timeout) or [it.each](https://jestjs.io/docs/en/api#testeachtablename-fn-timeout).

client/src/services/formulas/evaluator.test.js
```diff
@@ -67,4 +68,44 @@ describe("Evaluations", () => {
   it("can evaluate an expression with an NotEqual operator", () => {
     expect(evaluate("1 <> 2").value).toBe(true);
   });
+
+  it("can evaluate a boolean literal", () => {
+    expect(evaluate("TRUE").value).toBe(true);
+    expect(evaluate("FALSE").value).toBe(false);
+  });
+
+  it("can evaluate AND statements", () => {
```
marcja, line 77: Stylistically, this tests should be named "evaluates AND statements". The "can" is repetitive and unneeded.

client/src/services/formulas/evaluator.test.js
marcja: If you have a test elsewhere that proves that "1 < 2 = TRUE" or "IS_STUDENT = TRUE", what's the point of adding those tests here?

client/src/services/formulas/evaluator.test.js
```diff
@@ -67,4 +68,44 @@ describe("Evaluations", () => {
...
+  });
+
+  it("can evaluate IF statements", () => {
+    expect(evaluate("IF(TRUE,FALSE,TRUE)").value).toBe(false);
+  });
+
+  it("can evaluate IFS statements", () => {
+    expect(evaluate("IFS(TRUE,FALSE,FALSE,TRUE)").value).toBe(false);
+    expect(evaluate("IFS(FALSE,FALSE,TRUE,TRUE)").value).toBe(true);
+  });
```
marcja, line 110: Add tests for error cases, such as "AND()".

client/src/services/formulas/lexer.js
```diff
@@ -51,10 +51,9 @@ const LParen = createToken({ name: "LParen", pattern: /\(/ });
 const RParen = createToken({ name: "RParen", pattern: /\)/ });
 const NumberLiteral = createToken({
   name: "NumberLiteral",
-  pattern: /[1-9]\d*/,
+  pattern: /[0-9]\d*/,
```
marcja, line 54: The regular expression from that paper I sent is `/[0-9]+\.?[0-9]*/`. Note that number literals don't have to be integers.

client/src/services/formulas/lexer.js
```diff
@@ -93,6 +92,59 @@ const NotEqual = createToken({
...
 
+const BooleanLiteral = createToken({
+  name: "BooleanLiteral",
+  pattern: /(TRUE|FALSE)/,
+});
+
+// ------------------ FUNCTIONS ------------------
+const Function = createToken({
+  name: "Function",
+  pattern: Lexer.NA,
```
marcja, line 103: It's interesting to note that that paper I sent includes the left-paren in the matcher. So it matches "AND(", not just "AND". I'm assuming that it makes it easier to distinguish between functions and variables.

client/src/services/formulas/lexer.js
```diff
@@ -93,6 +92,59 @@ const NotEqual = createToken({
...
+});
+const FunctionName = createToken({
+  name: "FunctionName",
+  pattern: Lexer.NA,
+});
```
marcja, line 108: Since each function name is a reserved word from a finite set, why not lex them as distinct tokens?

client/src/services/formulas/lexer.js
```diff
@@ -93,6 +92,59 @@ const NotEqual = createToken({
...
+  categories: FunctionName,
+});
+const Ifs = createToken({
+  name: "Ifs",
+  pattern: /IFS/,
+  categories: FunctionName,
+});
+const If = createToken({
+  name: "If",
+  pattern: /IF/,
```
marcja, line 144: Is this guaranteed not to match the beginning of the IFS token?

client/src/services/formulas/parser.js
```diff
@@ -65,7 +73,7 @@ class Evaluator extends CstParser {
         // in the "lowest" leaf in the expression ParseTree.
         { ALT: () => $.SUBRULE($.parenthesisExpression) },
         { ALT: () => $.CONSUME(NumberLiteral) },
-        { ALT: () => $.SUBRULE($.powerFunction) },
+        { ALT: () => $.SUBRULE($.function) },
```
marcja, line 76: I think I'm not yet understanding the value of this refactor....

client/src/services/formulas/actions.js
tariqajyusuf: That's a much easier implementation lol.

client/src/services/formulas/evaluator.test.js
tariqajyusuf: It probably makes sense to test the ordering of operations.

client/src/services/formulas/lexer.js
tariqajyusuf: You're right, forgot to clean that up.

client/src/services/formulas/lexer.js
tariqajyusuf: I think that makes sense, but part of me doesn't like the idea of having to specify parentheses every time mainly since we already defined parens.

client/src/services/formulas/lexer.js
tariqajyusuf: The main reason of having the category here makes this a bit easier to distinguish between identifiers and functions and is in preparation for the dispatcher pattern I mentioned above.

client/src/services/formulas/lexer.js
tariqajyusuf: Yes as per the tests. Ordering here also matters and so this in fact fails to build if the order is reversed.

client/src/services/formulas/parser.js
tariqajyusuf: See dispatcher comment above

client/src/services/formulas/lexer.js
```diff
@@ -143,7 +199,11 @@ export default {
     const lexingResult = EvaluatorLexer.tokenize(inputText);
 
     if (lexingResult.errors.length > 0) {
-      throw Error("Lexing errors detected");
+      var errorMessage = "Errors were found while parsing:\n";
```
marcja, line 202: TODO: make formula errors more readable to a non-technical audience

PR #186, opened 2020-07-30
Update summary page

mscharfstein:
- Print out the questions and their responses by section
- Do not include questions with boolean responses of "No"
- Do not include sections if all of the questions within the section have a "No" response
- Print numbers with appropriate commas and currency with '$'
- Add edit buttons to each section (buttons do not yet do anything)
- Add start over button which sends user back to beginning and clears state (update responsesSlice to have a new clearState action)

image

TODO: Update edit buttons to send back to right part of the form
TODO: Send user to Services upon clicking SEE MY SERVICES


client/src/components/SummarySection.jsx
```diff
@@ -0,0 +1,88 @@
...
+                return (
+                  <li key={questionKey}>
+                    {t(`catalog:${questionKey}.text`)} {t(`responses.true`)}
+                  </li>
+                );
+              } else if (Number(responses[questionKey])) {
+                return (
+                  <li key={questionKey}>
+                    {t(`catalog:${questionKey}.text`)}{" "}
+                    {displayNumber(questionKey)}
```
marcja, line 69: This block is suspicious similarly to the previous block. Maybe extracting to a function will make it more [DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself).

client/src/components/SummarySection.jsx
```diff
@@ -0,0 +1,88 @@
...
+                  </li>
+                );
+              } else {
+                return <></>;
```
marcja: Is this needed?

mscharfstein: i changed it to "" but yes, otherwise a warning is thrown. you have to return something from the map

client/src/components/SummarySection.jsx
```diff
@@ -0,0 +1,88 @@
...
+          </Grid>
+          <ul>
+            {questionKeys.map((questionKey) => {
+              if (responses[questionKey] === "true") {
+                return (
+                  <li key={questionKey}>
+                    {t(`catalog:${questionKey}.text`)} {t(`responses.true`)}
+                  </li>
+                );
+              } else if (Number(responses[questionKey])) {
```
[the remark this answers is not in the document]
mscharfstein: yes, this is just to check whether we are dealing with an integer or a boolean. i could use the questiontype instead if you prefer

client/src/components/SummarySection.jsx
```diff
@@ -0,0 +1,88 @@
...
+                <Header as="h3">{t(`catalog:${sectionKey}.title`)}</Header>
+              </Grid.Column>
+              <Grid.Column width={3}>
+                <Link to="/">{t(`form.edit`)}</Link>
+              </Grid.Column>
+            </Grid.Row>
+          </Grid>
+          <ul>
+            {questionKeys.map((questionKey) => {
+              if (responses[questionKey] === "true") {
```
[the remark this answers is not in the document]
mscharfstein: it actually seems to be stored as "true" so it does need to be checked

client/src/components/SummaryPage.jsx
```diff
@@ -12,19 +15,42 @@ import { selectResponses } from "../features/responses/responsesSlice";
...
+          return (
+            <SummarySection
+              key={sectionKey}
+              sectionKey={sectionKey}
+              questionKeys={questionKeys}
+            />
+          );
+        });
+      })}
+      <div>
```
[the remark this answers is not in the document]
mscharfstein: whats the benefit of nav over div? updated though

client/src/components/SummaryPage.jsx
```diff
@@ -12,19 +15,41 @@ import { selectResponses } from "../features/responses/responsesSlice";
...
+        <Link to="/">
+          <Button primary>{t("button.services")}</Button>
```
itssumertime, line 41: Is this supposed to link to the Services page I'm adding?

client/src/components/SummarySection.jsx
```diff
@@ -0,0 +1,88 @@
...
+    const questionMetadata = questions[questionKey];
+    let result = formatNumber(responses[questionKey]);
+    if (questionMetadata.type === "CURRENCY") {
+      result = "$" + result;
+    }
+    return result;
+  };
+
+  const formatNumber = (num) => {
+    return num.toString().replace(/(\d)(?=(\d{3})+(?!\d))/g, "$1,");
```
itssumertime, line 40: What type of numbers is this formatting?? Could you also add a comment to this function please to indicate what it's meant to do?

client/src/components/SummaryPage.jsx
```diff
@@ -12,19 +15,42 @@ import { selectResponses } from "../features/responses/responsesSlice";
...
-      <h1>{t("summary.title")}</h1>
-      {Object.entries(responses).map(([questionKey, response]) => (
-        <p key={questionKey}>
-          {questionKey}: {String(response)}
-        </p>
-      ))}
-      <Link to="/">
-        <Button>{t("button.back")}</Button>
-      </Link>
+    <Container text as="main">
```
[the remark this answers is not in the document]
mscharfstein: thanks for the callout - it is a direct descendent

client/src/components/SummarySection.jsx
```diff
@@ -0,0 +1,88 @@
...
+              <Grid.Column width={13}>
+                <Header as="h3">{t(`catalog:${sectionKey}.title`)}</Header>
+              </Grid.Column>
+              <Grid.Column width={3}>
+                <Link to="/">{t(`form.edit`)}</Link>
+              </Grid.Column>
+            </Grid.Row>
+          </Grid>
+          <ul>
+            {questionKeys.map((questionKey) => {
```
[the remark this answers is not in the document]
mscharfstein: done

client/src/components/SummarySection.jsx
mscharfstein: done

client/src/components/SummarySection.jsx
mscharfstein: removed this and replaced with an existing function

client/src/components/SummaryPage.jsx
mscharfstein: yes

client/src/components/SummaryPage.jsx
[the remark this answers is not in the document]
marcja: <nav> is a semantic tag that means something to screenreaders. <div> is not.

client/src/components/SummarySection.jsx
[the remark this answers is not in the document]
marcja: That seems like a bug. Why aren't we storing it in redux as `true`?

client/src/components/SummarySection.jsx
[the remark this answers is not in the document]
marcja: yes

client/src/components/SummarySection.jsx
marcja: Ah

client/src/components/SummarySection.jsx
```diff
@@ -0,0 +1,88 @@
...
+/**
+ * Displays the question responses submitted by the user for a specific section of questions.
+ *
+ * @returns {React.ReactElement} div of question responses
+ */
+function SummarySection(props) {
+  const { sectionKey, questionKeys } = props;
+  const { t } = useTranslation();
+  const responses = useSelector(selectResponses);
+  const [showSection, setSection] = useState(false);
```
kebrahim, line 19: `setShowSection` for clarity?

client/src/components/SummarySection.jsx
```diff
@@ -0,0 +1,88 @@
...
+  const questions = useSelector(selectQuestions);
+
+  useEffect(() => {
+    questionKeys.forEach((questionKey) => {
+      if (responses[questionKey] === "true" || Number(responses[questionKey])) {
```
kebrahim, line 24: do we need to validate that the number questions have a valid positive #? isn't there a case where the user enters an invalid number (e.g. -1), it gets saved to Redux & then they navigate to this page using the URL?

client/src/components/SummarySection.jsx
kebrahim, line 40: does this need to be i18n-ized?

client/src/components/SummarySection.test.js
```diff
@@ -0,0 +1,167 @@
...
+      });
+      expect(heading).toBeInTheDocument();
+    });
+
+    it("renders edit link", async () => {
+      const editLink = screen.getByRole("link", {
+        name: "form.edit",
+      });
+      expect(editLink).toBeInTheDocument();
+      // TODO(mscharfstein): once edit link actually works, test its navigation
```
kebrahim, line 68: remove the "(mscharfstein)"; our TODO workflow doesn't handle it well

client/src/components/SummarySection.test.js
```diff
@@ -0,0 +1,167 @@
...
+      const editLink = screen.getByRole("link", {
+        name: "form.edit",
+      });
+      expect(editLink).toHaveFocus();
+      userEvent.tab(); // onto body
+      expect(document.body).toHaveFocus();
+    });
+  });
+  describe("empty summary section", () => {
+    it("section not displayed", () => {
```
kebrahim, line 102: s/section/is

client/src/components/SummaryPage.test.js
```diff
@@ -0,0 +1,123 @@
...
+    const responseTwo = screen.getByText(/4/);
+    expect(responseTwo).toBeInTheDocument();
+    const questionThree = screen.getByText(/HOUSEHOLD_INCOME.text/);
+    expect(questionThree).toBeInTheDocument();
+    const responseThree = screen.getByText(/\$1,000/);
+    expect(responseThree).toBeInTheDocument();
+  });
+});
+
+test("has no accessibility violations", async () => {
```
kebrahim, line 98: mind also adding a test for keyboard navigation?

client/src/components/SummarySection.jsx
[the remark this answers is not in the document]
mscharfstein: that i'm not sure - would be in the BooleanQuestion component. can file a bug to handle that?

client/src/components/SummarySection.jsx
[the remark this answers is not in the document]
mscharfstein: done

client/src/components/SummarySection.jsx
mscharfstein: done

client/src/components/SummarySection.jsx
mscharfstein: hm, if they enter an invalid number they will not be able to continue or submit the form

if we think they could navigate to this page even having not submitted the form, we should do additional checks that all the responses are recorded, otherwise the page will break

client/src/components/SummarySection.jsx
mscharfstein: not using this anymore, but i think we said everything can be in US dollars

client/src/components/SummaryPage.jsx
```diff
@@ -17,11 +17,16 @@ function SummaryPage() {
...
   const form = useSelector(selectForm);
   const dispatch = useDispatch();
+  const history = useHistory();
 
   const handleStartOver = () => {
     dispatch(clearResponses());
   };
 
+  const handleServicesClick = () => {
+    history.push("/"); // TODO: Link button to services page
```
marcja, line 27: Push to "/app/questions" instead. See Routes.jsx.

client/src/components/SummaryPage.jsx
```diff
@@ -36,20 +41,20 @@ function SummaryPage() {
...
-          <Button primary>{t("button.services")}</Button>
-        </Link>
-      </div>
-      <div>
+      <nav aria-label={t("button.services")}>
+        <Button primary onClick={handleServicesClick}>
+          {t("button.services")}
+        </Button>
+      </nav>
+      <nav>
```
marcja, line 49: As I said before, there should be one `<nav>` including both buttons, not two separate ones. It represents the navigation section of the page.

client/src/components/SummaryPage.test.js
```diff
@@ -27,11 +28,11 @@ describe("summary page", () => {
...
         HOUSEHOLD_SIZE: 4,
-        ADULTS_65_PLUS: true,
-        CHILDREN_00_05: false,
+        ADULTS_65_PLUS: "true",
+        CHILDREN_00_05: "false",
         HOUSEHOLD_INCOME: 1000,
-        IS_PREGNANT: true,
-        IS_EMPLOYED: false,
+        IS_PREGNANT: "true",
+        IS_EMPLOYED: "false",
```
marcja, line 35: Again, this is clearly a bug in how we are storing responses.

client/src/components/SummarySection.jsx
```diff
@@ -16,28 +16,50 @@ function SummarySection(props) {
...
       }
     });
-  }, [responses, questionKeys]);
+  }, [responses, questionKeys, questions]);
 
   const displayNumber = (questionKey) => {
     const questionMetadata = questions[questionKey];
-    let result = formatNumber(responses[questionKey]);
+    // Format number based on US standard to include commas in appropriate places
+    let result = new Intl.NumberFormat("en-US").format(responses[questionKey]);
```
marcja, line 38: :| Intl.NumberFormat supports formatting currencies. It's literally the first example on the documentation page. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/NumberFormat

```js
new Intl.NumberFormat("en-US", { style: 'currency', currency: 'USD' }).format(...)
```

client/src/components/SummaryPage.test.js
```diff
@@ -0,0 +1,123 @@
...
+    expect(startOverLink).toBeInTheDocument();
+  });
+
+  it("has correct questions and responses", async () => {
+    expect(screen.getAllByRole("listitem")).toHaveLength(3);
+    expect(
+      screen.getAllByRole("link", {
+        name: "form.edit",
+      })
+    ).toHaveLength(2);
```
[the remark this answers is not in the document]
mscharfstein: listitems are just the bullets - they don't have links. only sections have edit links. there are 2 sections with 3 bullets across both of them. added comments

client/src/components/SummaryPage.jsx
mscharfstein: this is just a placeholder - it is actually going to go to the Services page, which I assume will be at app/services or app/report?

client/src/components/SummaryPage.jsx
mscharfstein: didn't see before, sorry. thanks

client/src/components/SummaryPage.test.js
mscharfstein: i know, i filed a bug for it and i see where the bug is (in BooleanQuestion), but I don't want to fix it in this PR

client/src/components/SummarySection.jsx
mscharfstein: i don't want to format the currency on this line because not all number responses are currency (some are just numbers). since i've already formatted it to a number, it's easier to add a $ then use Intl.NumberFormat again

client/src/components/SummaryPage.jsx
marcja: the route is defined in a PR that @itssumertime is working on and not yet merged

client/src/components/SummaryPage.jsx
```diff
@@ -12,19 +15,44 @@ import { selectResponses } from "../features/responses/responsesSlice";
...
+              questionKeys={questionKeys}
+            />
+          );
+        });
+      })}
+      <nav>
+        <Button primary onClick={handleServicesClick}>
+          {t("button.services")}
+        </Button>
+        <Link
```
marcja, line 48: Does this render the way you want it to? Button and Link are both inline elements, so they are probably appearing next to each other....

client/src/components/SummaryPage.test.js
marcja: k

client/src/components/SummaryPage.test.js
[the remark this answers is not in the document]
marcja: Fine for now, but this test should be cleaned up in the future. Reading the test right now doesn't help me understand the desired functionality very well.

client/src/components/SummarySection.jsx
marcja: Really?
```js
// at module scope
const numberFormatter = new Intl.NumberFormat("en-US");
const currencyFormatter = new Intl.NumberFormat("en-US"), { style: "currency", currency: "USD" });
// here in this code
return questionMetaData.type === "CURRENCY" ?
  currencyFormatter(responses[questionKey]) :  
  numberFormatter(responses[questionKey]);
```

client/src/components/SummarySection.jsx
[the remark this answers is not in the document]
mscharfstein: in the meantime, i have submitted a fix for that bug and have merged it, so updated how this is handled

client/src/components/SummaryPage.jsx
mscharfstein: wanted to wait for that to be completed. updated now

client/src/components/SummaryPage.jsx
mscharfstein: putting these in the same nav make them render on the same line. i'm not sure the best way to handle this - i surrounded each by a div

client/src/components/SummaryPage.jsx
mscharfstein: yeah i just replied that in response to your original comment about putting them in the same nav, i added divs. if there's a different way to handle with css, i can leave it for a separate PR

PR #187, opened 2020-07-30
Added instructions field to JSON

kebrahim:
Works on #142 

Note that I made the `instructions` field able to be blank, as I noticed that the Victim Compensation service doesn't have any application instructions.

Before PR: 
instructions_before

After PR: 
instructions_after


server/src/main/java/gov/ny/its/hs/maslow/model/ServiceResources.java
```diff
@@ -18,6 +18,7 @@
   @NotBlank
   private String description;
 
+  private String instructions;
```
marcja, line 21: Don't you also need to add this to the DTO?

PR #237, opened 2020-07-30
Factor out functions to a separate file

tariqajyusuf:
More work on #163 

Refactors functions to an external file to make it a bit more readable.


tariqajyusuf:
@mscharfstein it's hard to distinguish between logic for operators and ordering compared to functions. I'd argue that for maintainability it's much easier to add more functions by having it factored out separately.

client/src/services/formulas/evaluator.test.js
```diff
@@ -79,34 +80,41 @@ describe("Evaluations", () => {
     expect(evaluate("AND(TRUE,TRUE)").value).toBe(true);
     expect(evaluate("AND(TRUE,FALSE,TRUE)").value).toBe(false);
     expect(evaluate("AND(TRUE,TRUE,TRUE)").value).toBe(true);
+    expect(() => evaluate("AND(TRUE)").value).toThrow();
```
marcja, line 83: Why does this throw? This (surprisingly) valid in spreadsheet formulas.

client/src/services/formulas/evaluator.test.js
```diff
@@ -79,34 +80,41 @@ describe("Evaluations", () => {
...
   });
 
   it("evaluate OR statements", () => {
     expect(evaluate("OR(TRUE,FALSE)").value).toBe(true);
     expect(evaluate("OR(FALSE,FALSE)").value).toBe(false);
     expect(evaluate("OR(TRUE,FALSE,FALSE)").value).toBe(true);
     expect(evaluate("OR(FALSE,FALSE,TRUE)").value).toBe(true);
+    expect(() => evaluate("OR(TRUE)").value).toThrow();
```
marcja, line 91: Same here.

client/src/services/formulas/evaluator.test.js
```diff
@@ -79,34 +80,41 @@ describe("Evaluations", () => {
...
   });
 
   it("evaluate XOR statements", () => {
     expect(evaluate("XOR(TRUE,FALSE)").value).toBe(true);
     expect(evaluate("XOR(TRUE,TRUE)").value).toBe(false);
     expect(evaluate("XOR(TRUE,FALSE,FALSE)").value).toBe(true);
     expect(evaluate("XOR(FALSE,FALSE,TRUE)").value).toBe(true);
+    expect(() => evaluate("XOR(TRUE)").value).toThrow();
```
marcja, line 99: Same here.

client/src/services/formulas/actions.js
```diff
@@ -4,6 +4,7 @@
 
 import lexer from "./lexer";
 import parser from "./parser";
+import dispatch from "./functions";
```
marcja, line 7: This is an unfortunate name collision, given that `dispatch` is used all over the codebase specifically for redux. Please rename to something else.

client/src/services/formulas/functions.js
```diff
@@ -0,0 +1,101 @@
...
+      "Expected at least ",
+      minParamCount,
+      " arguments but got ",
+      params.length
+    );
+    throw Error(message);
+  }
+};
+
+const power = (params) => {
```
marcja, line 20: Suggest renaming all of these constants to uppercase, like `POWER`, `AND`, `OR`.

client/src/services/formulas/functions.js
```diff
@@ -0,0 +1,101 @@
...
+  return Math.pow(params[0], params[1]);
+};
+
+const and = (params) => {
+  let result = params[0];
+
+  for (const param of params) {
```
marcja, line 27: Why is the loop for AND different than the loop for OR? Shouldn't they only differ by `&&` vs `||`?

client/src/services/formulas/functions.js
```diff
@@ -0,0 +1,101 @@
...
+  }
+
+  return !!(numTrues % 2);
+};
+
+const not = (params) => {
+  return !params[0];
+};
+
+const ifFunc = (params) => {
```
marcja, line 58: Uppercase would solve your problem here: IF

marcja: My other thought would be to put these in an object:
```js
const Functions = {
  AND: (params) => { return ...; },
  NOT: (params) => { return ...; },
  IF: (params) => { return ...; },
};
```
and reference below as:
```js
const dispatcher = [
  [PowerFunc, Functions.POWER, 2],
  ...
];
```

client/src/services/formulas/functions.js
```diff
@@ -0,0 +1,101 @@
...
+    if (params[i]) {
+      return params[i + 1];
+    }
+  }
+};
+
+// Format is [token, function, minimum arguments].
+const dispatcher = [
+  [PowerFunc, power, 2],
+  [And, and, 2],
```
marcja, line 83: AND, OR, and XOR all allow a single argument, believe it or not. Try it out in Sheets.

client/src/services/formulas/functions.js
```diff
@@ -0,0 +1,101 @@
...
+  [Or, or, 2],
+  [Not, not, 1],
+  [Xor, xor, 2],
+  [If, ifFunc, 3],
+  [Ifs, ifs, 2],
+];
+
+const dispatch = (functionName, params) => {
```
marcja, line 91: Please rename.

client/src/services/formulas/functions.js
```diff
@@ -0,0 +1,101 @@
...
+  // TODO: Check on param count and throw errors if the count is invalid.
+  for (let option of dispatcher) {
```
marcja, line 93: `s/let/const/`

client/src/services/formulas/functions.js
```diff
@@ -0,0 +1,101 @@
...
+    if (tokenMatcher(functionName, option[0])) {
+      checkParams(option[0].tokenName, params, option[2]);
+      return option[1](params);
+    }
+  }
```
marcja, line 98: FWIW:
```js
const match = dispatcher.find((option) => tokenMatcher(functionName, option[0]));
return checkParams(match[0].tokenName, params, match[2]) && option[1](params);
```

PR #238, opened 2020-07-30
Adjust links on Summary page to send you back to the correct Section

mscharfstein:
- Passes state (page number) to "/app/questions" so that it can render the right section
- Works on localhost, but cannot get the test to pass - another set of eyes would be really helpful :)


client/src/components/SummarySection.test.js
```diff
@@ -65,7 +77,8 @@ describe("summary section", () => {
         name: "form.edit",
       });
       expect(editLink).toBeInTheDocument();
-      // TODO: once edit link actually works, test its navigation
+      fireEvent.click(editLink);
+      //await waitFor(() => screen.getByTestId("formStepMain"));
```
mscharfstein, line 81: seems like this should work but it doesnt

kebrahim: maybe it's because you're using the wrong `waitFor` method? you should be using the one in testing-library/react, not testing-library/dom - https://github.com/ny/maslow/blob/master/client/src/App.test.js#L1

what error do you get?

client/src/components/SummarySection.jsx
```diff
@@ -72,7 +83,17 @@ function SummarySection(props) {
                 <Header as="h3">{t(`catalog:${sectionKey}.title`)}</Header>
               </Grid.Column>
               <Grid.Column width={3}>
-                <Link to="/">{t(`form.edit`)}</Link>
+                <Link
+                  to={{
+                    pathname: "/app/questions",
+                    state: {
+                      page: pageNumberFromSection(sectionKey),
+                      showError: true,
```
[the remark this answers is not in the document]
mscharfstein: since they've already filled out the form, i think we should show an error as soon as you go to the page (granted, there will be no errors at first because they successfully submitted)

client/src/components/SummarySection.test.js
mscharfstein: oh good thought!! but i don't get an error with the specific function, it just says it can't find the element (Unable to find an element by: [data-testid="formStepMain"]). seems like it's not actually getting to the right page (though it works on localhost) or its not waiting for the page to render correctly. updated the library and still the same thing

client/src/components/SummarySection.jsx
```diff
@@ -72,7 +83,17 @@ function SummarySection(props) {
                 <Header as="h3">{t(`catalog:${sectionKey}.title`)}</Header>
               </Grid.Column>
               <Grid.Column width={3}>
-                <Link to="/">{t(`form.edit`)}</Link>
+                <Link
+                  to={{
```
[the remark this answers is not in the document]
mscharfstein: that's just how it gets formatted.. but one set of brackets is to indicate its a variable within the rendering code. the other is the object

client/src/components/SummarySection.test.js
kebrahim: i did a little research. it looks like what you're trying to do may not be possible: https://stackoverflow.com/questions/57827126/how-to-test-anchors-href-with-react-testing-library

to that point, i don't think we really need to test that clicking on an anchor tag navigates the user to the corresponding href. i think that's pretty well tested ;)

i think it's sufficient to verify that the href of the anchor tag is the URL that you expect it to be. however, is it also possible to check that the `state` is set to the page number you expect?

PR #240, opened 2020-07-30
Store responses to boolean questions as booleans (as opposed to strings) in redux

### LINKED ISSUE #208: Update answers Redux storage to store BOOLEAN true/false vs. strings
- Seems like Redux store has strings "true" / "false" rather than BOOLEANs
- Once fixed, SectionSummary can be updated where it looks for the answers to be true

mscharfstein:
Fixes #208 

- Have to specifically set the value as a boolean when updating the state (updated the "value" of the Form field to be a boolean instead of a string, but that wasn't enough)


client/src/components/BooleanQuestion.jsx
```diff
@@ -36,9 +36,13 @@ function BooleanQuestion(props) {
   };
 
   const handleChange = (event) => {
+    let response = false;
+    if (event.target.value === "true") {
+      response = true;
+    }
```
egracer, line 42: I think this could just be `const response = (event.target.value === 'true');`.

client/src/components/BooleanQuestion.jsx
```diff
@@ -79,9 +83,9 @@ function BooleanQuestion(props) {
           className={radioItemClass}
           label={t("radio.yes")}
           id={`radio-yes-${questionKey}`}
-          value="true"
+          value={true}
```
egracer, line 86: I distinctly remember an error in the console suggesting that `value` had to be a string. I originally did it this way but reverted because of that console error. Maybe something has changed since then -- this is fine by me as long as we're not introducing errors, all tests, and the toggle works.

mscharfstein: ack, everything still works and not seeing any errors. if they start popping up, I can figure out a way to get around this (harder because its a controlled component so I don't want to leave this as a string if it will get a boolean from redux)

client/src/components/BooleanQuestion.jsx
mscharfstein: done

PR #246, opened 2020-07-30
Remove unnecessary todo

### LINKED ISSUE #197: Change tests to start with a leading '=' once the
Change tests to start with a leading '=' once the

parser can handle it.

https://github.com/ny/maslow/blob/f32f71874b0613c262e4b1fbd7bb9518484fcf9c/client/src/services/formulas/evaluator.test.js#L3

```javascript

// TODO: Change tests to start with a leading '=' once the
// parser can handle it.
describe("Evaluations", () => {
  it("can evaluate an expression", () => {
ndex aa04c7b..d3e8dbc 100644
++ b/client/src/services/formulas/lexer.js
```

9e6968ea5b7c505180c849bee26cfc047c43fdcf

mscharfstein:
Fixes #197 

Note that when reading from yaml, we will need to ignore the leading "=" when passing it in to the function that evaluates the string


client/src/services/formulas/evaluator.test.js
```diff
@@ -1,7 +1,5 @@
 import evaluate from "./actions";
 
-// TODO: Change tests to start with a leading '=' once the
```
kebrahim, line 3: should the TODO be moved to whichever class is going to be passing the formula into the parser?

alternatively, should we remove the `=` from the yaml?

mscharfstein: i think we should just remove = from yaml but there's an argument to be made that keeping "=" makes it clear its an excel formula. added a todo in the services slice because that's where we will call the evaluator

kebrahim: i don't see the change to servicesSlice in this PR.

also, if we do want to propose making a change like that to the yaml, we'll have to run it by james

mscharfstein: pushed now

